PR #33, opened 2016-11-16
Port HTML's example of algorithm steps implying must

domenic:
Not sure if we should say anything about how "the user" is implied by context. Usually it's "the user agent"... Maybe we should redo the example.


annevk:
I think it's fine, but we do need IDs for Bikeshed.

PR #34, opened 2016-11-16
Tweak some of the definitions for maps

domenic:
This makes some changes to the map verbiage which were discovered to be helpful while working on https://github.com/whatwg/html/pull/2054. Concretely, it:

- Adds an explicit concept of getting
- Makes the shorthands for set/remove/exists link to the longhands, instead of defining a second similar term
- Changes the IDs and linking texts of the concepts to be shorter and more useful

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -333,15 +333,15 @@ An indexing syntax can be used to look up and set values by providing a key insi
 
 <p>To <dfn export for=map lt="get|get the value">get the value of an entry</dfn> in an
 <a>ordered map</a> given a key is to retrieve the value of any existing key/value pair if the map
-<a lt="contains an entry with a given key">contains an entry with the given key</a>, or if to return
-nothing otherwise. We can also use the indexing syntax explained above.
+<a for=map>contains</a> an entry with the given key, or if to return nothing otherwise. We can also
+use the indexing syntax explained above.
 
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -331,21 +331,26 @@ An indexing syntax can be used to look up and set values by providing a key insi
...
+all entries from the map that match a given condition, or do nothing if none do. If the condition is
+having a certain key, then we can also denote this by saying, for an <a>ordered map</a> |map| and
+key |key|, "<a for=map>remove</a> |map|[|key|]".
 
-<p>An <a>ordered map</a> <dfn export for=map>contains an entry with a given key</dfn> if there
-exists a key/value pair with that key. We can also denote this by saying that, for an
-<a>ordered map</a> |map| and key |key|, "|map|[|key|] <dfn export for=map>exists</dfn>".
+<p>An <a>ordered map</a> <dfn export for=map lt=exists|contains>contains an entry with a given
+key</dfn> if there exists a key/value pair with that key. We can also denote this by saying that,
+for an <a>ordered map</a> |map| and key |key|, "|map|[|key|] <a for=map>exists</dfn>".
```
annevk: End tag here is wrong. I guess I can fix that.

annevk:
Found another issue when running make, fixed all I think, but would appreciate another look.

domenic:
Thanks. I should really be running make instead of make remote.

Issue #22, opened 2016-11-10
Control flow for loops


annevk:
It would be great to have a less awkward way of phrasing to run a loop for the next item: https://dom.spec.whatwg.org/#concept-event-listener-inner-invoke. Basically something like "continue".

mikewest:
CSP uses "continue to the next item" in various places (https://w3c.github.io/webappsec-csp/#parse-serialized-policy for example). Since that's the way it's phrased in a variety of programming languages, phrasing it that way in the spec makes sense to me.

I'd suggest a template like the following:

> 1. For each |item| in |list|:
>     1. If we should skip |item| for some reason, <a>continue</a> to the next |item|.
>     2. If we should stop the loop entirely, <a>break</a>.
>     3. If we should skip out of this loop to a labeled step, <a>continue</a> to `label`.
>     4. Do a thing to |item|.

annevk:
I like "continue to label". I wonder if we can just use "continue"  for the other kind, since there's no real reason to make that phrase more complicated and mentioning the variable there makes it seem like you're passing an argument (which is a thing we do with "return" and "throw" and "continue to").

And maybe we should have break as well?

> If X, then continue.
> If Y, then continue to Z.
> If AA, then break.

mikewest:
SGTM.

annevk:
#56 discusses labels as well so I think we can leave that there for now and consider this fixed by @mikewest. (I'll post a PR for DOM to make use of the new concepts and get rid of the old language.)

PR #35, opened 2016-11-18
Define 'continue' and 'break' statements.

mikewest:
This patch defines 'continue' and 'break' in the context of a 'for each' loop, and
provides simple examples of their usage.

This addresses pieces of https://github.com/whatwg/infra/issues/22, leaving
'continue to' for a subsequent patch.

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -107,55 +107,59 @@ leading spaces" or "return false") are to be interpreted with the meaning of the
 manner, so long as the end result is equivalent. (In particular, the algorithms are intended to be
 easy to follow, and not intended to be performant.)
 
+<h4 id=algorithm-control-flow>Control flow</h4>
+
 <p>The control flow of algorithms is such that a requirement to "return" or "throw" terminates the
 algorithm the statement was in. "Return" will hand the given value, if any, to its caller. "Throw"
 will make the caller automatically rethrow the given value, if any, and thereby terminate the
 caller's algorithm. Using prose the caller has the ability to "catch" the exception and perform
 another action.
 
-<p>An iteration's flow can be controlled via requirements to "continue" or "break". "Continue" will
-skip over any remaining steps in an iteration, proceeding to the next item. If no further items
-remain, the iteration will stop. "Break" will skip over any remaining steps in an iteration, and
-skip over any remaining items as well, stopping the iteration.
+<p>An iteration's flow can be controlled via requirements to
+<dfn export for=iteration>continue</dfn> or <dfn export for=iteration> break</dfn>.
+<a for=iteration>Continue</a> will skip over any remaining steps in an iteration, proceeding to the
+next item. If no further items remain, the iteration will stop. <a for=iteration>Break</a> will skip
+over any remaining steps in an iteration, and skip over any remaining items as well, stopping the
+iteration.
 
 <div class=example id=example-break-continue>
-  <p>Let |example| be the <a>list</a> « |1|, |2|, |3|, |4| ». The following prose would perform
-  <code>operation</code> upon |1|, then |2|, then |3|, then |4|:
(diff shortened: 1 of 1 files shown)


mikewest:
WDYT, @annevk? I think it might be worthwhile to pull the iteration discussion out into a "Control Flow" section; I didn't do that in this patch, but I think it's worth doing before trying to define "continue to".

annevk:
So I think rather than tying this to "for each" (of lists) I think we want to put this in the "Algorithms" section and leave it a little vaguer, but making it apply to all kinds of loops. "While", "for each", etc.

So group this together with where we define what "return" and "throw" result in.

As we flush that out more as time goes on it might warrant subsections indeed.

mikewest:
I took another pass at it.

Should "return", "throw", "continue", and "break" be wrapped in `<dfn>`? `abstract-op` perhaps?

annevk:
Looks good. Would anyone ever reference these terms?

mikewest:
No idea. *shrug* I suppose the indirection through "X depends on the Infra Standard" is enough, but `<a>continue</a>` would be more explicit.

annevk:
One problem is that IDL defines throw too. We need to figure out how to reconcile with IDL, but ideally not right now.

mikewest:
SGTM.

annevk:
Will let @domenic have a look too.

domenic:
I pushed some tweaks. I think it's important to define and export these things; specs might not want to always x-ref them, but they should at least have the option to do so.

I also grouped these and the return/throw thing under a heading, but I'm not 100% on that; let me know what you think.

Great examples, @mikewest.

mikewest:
Fixed a tiny typo in your fixups; otherwise LGTM. :)

infra.bs
```diff
@@ -107,12 +107,61 @@ leading spaces" or "return false") are to be interpreted with the meaning of the
...
+<h4 id=algorithm-control-flow>Control flow</h4>
+
 <p>The control flow of algorithms is such that a requirement to "return" or "throw" terminates the
 algorithm the statement was in. "Return" will hand the given value, if any, to its caller. "Throw"
 will make the caller automatically rethrow the given value, if any, and thereby terminate the
 caller's algorithm. Using prose the caller has the ability to "catch" the exception and perform
 another action.
 
+<p>An iteration's flow can be controlled via requirements to
+<dfn export for=iteration>continue</dfn> or <dfn export for=iteration>break</dfn>.
```
annevk: @domenic if we export these, why not export return and throw? Is the plan to do that later once we figure things out with IDL?

infra.bs
```diff
@@ -107,12 +107,61 @@ leading spaces" or "return false") are to be interpreted with the meaning of the
...
+ </ol>
+
+ <p>The following prose would perform <code>operation</code> upon 1, then 2, then 4. 3 would be
+ skipped.
+
+ <ol>
+  <li>
+   <p><a for=list>For each</a> |item| in |example|:
+   <ol>
+    <li>If |item| is 3, <a for=iteration>continue</a>.
```
annevk: the continue*

infra.bs
```diff
@@ -107,12 +107,61 @@ leading spaces" or "return false") are to be interpreted with the meaning of the
...
+ </ol>
+
+ <p>The following prose would perform <code>operation</code> upon 1, then 2. 3 and 4 would be
+ skipped.
+
+ <ol>
+  <li>
+   <p><a for=list>For each</a> |item| in |example|:
+   <ol>
+    <li>If |item| is 3, <a for=iteration>break</a>.
```
annevk: then break*

annevk:
I'm happy to fix the remaining editorial problems once we sort out the export story.

infra.bs
domenic: Yes, we should wait there.

annevk:
Fixed remaining editorial issues.

PR #38, opened 2016-11-30
Breaking: adjust some code point definitions

annevk:
Adjust some terms per feedback given in
https://github.com/whatwg/html/pull/2061. Downstream will have to
adjust to the new terms.


domenic:
LGTM, will let you merge so you can minimize the disruption to other specs.

Maybe we should create a list of dependent specs so we can file issues in the case of breaking changes? I'll create a separate PR for that.

TohaAbir (commented):
#461

PR #44, opened 2016-12-13
Exporting `append` and `prepend` for set

Sun77789:
@mikewest


domenic:
Thanks! Can you do prepend too, and add your name to the acknowledgments?

Sun77789:
Updated! Thank you :)

PR #45, opened 2016-12-16
Adding `lt=contain`

Sun77789:
@mikewest

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -331,8 +331,8 @@ from the list that match a given condition, or do nothing if none do.
  "<code>b</code>", "<code>ba</code>" ».
 </div>
 
-<p>A <a>list</a> <dfn export for=list,stack,queue,set>contains</dfn> an item if it appears in the
-list.
+<p>A <a>list</a> <dfn export for=list,stack,queue,set lt=contain>contains</dfn> an item if it 
+appears in the list.
 
 <p>A <a>list</a>'s <dfn export for=list,stack,queue,set>size</dfn> is the number of items the list
 <a for=list>contains</a>.


Sun77789:
@annevk

infra.bs
```diff
@@ -331,8 +331,8 @@ from the list that match a given condition, or do nothing if none do.
  "<code>b</code>", "<code>ba</code>" ».
 </div>
 
-<p>A <a>list</a> <dfn export for=list,stack,queue,set>contains</dfn> an item if it appears in the
-list.
+<p>A <a>list</a> <dfn export for=list,stack,queue,set lt="contain">contains</dfn> an item if it
```
mikewest: Nit: I think this doc generally avoids quotes when not necessary. So, probably prefer `lt=contain`.

Sun77789: Updated!

annevk:
How would you use this in a sentence?

mikewest:
Bikeshed converts `contain` to `contains` and `contained`, but doesn't convert `contains` to `contained` (see, for instance, step 4 of https://w3c.github.io/webappsec-csp/embedded/#intersection-policies). @tabatkins suggested that the right thing to do is change the `<dfn>` rather than adding even more logic to the variation machinery in Bikeshed: https://github.com/w3c/webappsec-csp/pull/169#issuecomment-267452007

annevk:
Yeah, I understand that, but I think what we defined here is only intended for usage of "If LIST contains X" and I couldn't find out how one would not write "contains" in such sentences.

Your usage seems a little confusing. You are trying to construct a list with it, which is not what it's for.

Additionally, by saying "either A or B" it's also not clear what the eventual order is going to be. Would it not make more sense to offer utilities to merge sets and such?

mikewest:
> You are trying to construct a list with it, which is not what it's for.

We're using `contained` as a conditional in some (admittedly vague) shorthand. If we rewrote that step to something like:

4. Let |names| be an empty list.
5. For each |directive| in |A|:
   1. Append |directive|'s name to |names|.
6. For each |directive| in |B|:
   1. Append |directive|'s name to |names| if |directive|'s name is not contained in |names|.

We'd still need "contained". I know we could rewrite that to something like "unless |names| contains |directive|'s name", but it seems strange to require specs to use a particular conjugation of a term.

> Additionally, by saying "either A or B" it's also not clear what the eventual order is going to be. Would it not make more sense to offer utilities to merge sets and such?

Indeed. For the particular use case here, that doesn't have practical impact, but I agree that we could clean it up.

annevk:
I guess that's reasonable. Will let @domenic make the final call.

PR #47, opened 2017-01-06
Editorial: generate the indices more correctly

domenic:
This was noted in https://github.com/whatwg/url/issues/190.


annevk:
LGTM, I guess. I don't really like acknowledgments not being the last thing in the document. Especially since we also picked that section to list the license and maintainer. Maybe we should move that to to its own section that can be truly at the end? Not coming up with any good names at the moment though and it would also require a change to the WHATWG Bikeshed template, but perhaps it means we can use the "Editor" line again.

domenic:
Hmm, I feel the opposite. To me, acknowledgments are way more important than indices, and so should appear earlier in the document.

annevk:
I can agree with acknowledgments, but the stuff tacked onto acknowledgments makes less sense there if it's not last, imo.

domenic:
Hmm, I see. I don't think that stuff is too crufty, personally, so I'm OK keeping it above the rest.

PR #48, opened 2017-01-07
Move string algorithms to Infra

shekyan:
Per https://github.com/whatwg/html/pull/2243#issuecomment-270594918,
attempting to move strictly splitting algorithm and some others to
this document. Not sure if I should only move strictly splitting, or
what I included in this PR, or more, inlcuding trimming, splitting on
spaces and commas, etc.


infra.bs
```diff
@@ -280,6 +280,81 @@ quotes and monospace font.
 <a>ASCII lowercase</a> of <var>B</var>.
 <!-- TODO: define string equals? -->
 
+To <dfn data-export="">collect a sequence of characters</dfn> means that the following algorithm
```
annevk: Missing a `<p>`.

I also think we want to rephrase this to make the arguments explicit.

(The other thing I'm a little unsure about is "characters", since that's not a term we currently use, but perhaps it's best for now since we haven't really figured out how we want to do all aspects of strings yet.)

infra.bs
```diff
@@ -280,6 +280,81 @@ quotes and monospace font.
...
+must be run, with <var>characters</var> being the set of characters that can be collected:</p>
+
+<ol>
+
```
annevk: This newline and the one before `</ol>`  can be removed. Same for the later algorithm.

infra.bs
```diff
@@ -280,6 +280,81 @@ quotes and monospace font.
...
+  <li><p>Let <var>input</var> and <var>position</var> be the same variables as
```
annevk: We use single-space indentation.

infra.bs
```diff
@@ -280,6 +280,81 @@ quotes and monospace font.
...
+
+<p>The step <dfn>skip whitespace</dfn> means that the user agent must <a>collect a sequence of
+characters</a> that are <a>ASCII whitespace</a> characters. <!--The step
+<dfn>skip White_Space characters</dfn> means that the user agent must <a>collect a sequence of
+characters</a> that are <span>White_Space</span> characters. In both cases, the--> The collected
+characters are not used. <!--<ref spec=UNICODE>--></p>
+
+<p>When a user agent is to <dfn data-export="">strip line breaks</dfn> from a string, the user
+agent must remove any U+000A LINE FEED (LF) and U+000D CARRIAGE RETURN (CR) characters from that
+string.</p>
```
annevk: We call it "newline", so "strip newlines" (perhaps with lt="strip newline" so you can also use the singular easily).

infra.bs
```diff
@@ -280,6 +280,81 @@ quotes and monospace font.
...
+
+<p>When a user agent is to <dfn data-export="">strip leading and trailing whitespace</dfn> from a
+string, the user agent must remove all <span data-x="space character">space characters</span> that
```
annevk: space characters is no longer a thing. This should be ASCII whitespace.

annevk: But I'm not sure we need to define this one. It seems "remove any leading and trailing ASCII whitespace from _str_" is equally short.

infra.bs
```diff
@@ -280,6 +280,81 @@ quotes and monospace font.
...
+are at the start or end of the string.</p>
+
+<p>When a user agent is to <dfn data-export="">strip and collapse whitespace</dfn> in a string, it
+must replace any sequence of one or more consecutive <span data-x="space character">space
+characters</span> in that string with a single U+0020 SPACE character, and then <a>strip
+leading and trailing whitespace</a> from that string.</p>
```
annevk: Space characters issue again.

annevk:
I forgot to say, I really appreciate your help getting this sorted.

infra.bs
domenic: I think it should be code points. That is what HTML says it means, at least: https://html.spec.whatwg.org/#character

infra.bs
domenic: Also, any data-x should no longer exist; that is a HTML-ism.

infra.bs
domenic: We should use just `export` in this document. I recognize you are porting from HTML, so sorry about the mismatch :)

infra.bs
```diff
@@ -280,6 +280,81 @@ quotes and monospace font.
 <a>ASCII lowercase</a> of <var>B</var>.
 <!-- TODO: define string equals? -->
 
+To <dfn data-export="">collect a sequence of characters</dfn> means that the following algorithm
+must be run, with <var>characters</var> being the set of characters that can be collected:</p>
```
domenic: Throughout the document try to follow the existing tag elision rules used here (but not used in HTML). Most close tags are omitted.

shekyan:
Thanks for all the comments. I am fine with fixing everything and will try address this week.

shekyan:
ping?

infra.bs
```diff
@@ -280,6 +280,61 @@ quotes and monospace font.
 <a>ASCII lowercase</a> of <var>B</var>.
 <!-- TODO: define string equals? -->
 
+<p>To <dfn export>collect a sequence of <a>code points</a></dfn> means that the following algorithm
+must be run:</p>
+<ol>
+<li><p>Let <var>characters</var> be the set of <a>code points</a> that can be collected.
```
annevk: It seems the indentation is still incorrect here. But I'm happy to reformat it for you.

annevk:
I mistakenly thought @domenic would pick it up. Sorry for dropping the ball. Shall I give it a pass and then you can review it @shekyan?

shekyan:
@annevk , sure, it'd be the best way to learn.

domenic:
Sorry, I don't think I got an email when the new commit was pushed! In general it's a good idea to ping the thread after doing so.

annevk:
@domenic I got stuck trying to rewrite these algorithms into "our style". The main problem is that the arguments are not explicit. I guess effectively these are functions that exist within a larger string parser class. Any ideas for how to phrase that or should we just stick to the language Ian came up with for HTML?

domenic:
@annevk I gave it a shot between BlinkOn talks. Probably needs some review and nit-fixing (not to mention rebasing), but I think this style works. It'll make calling the algorithms a bit more verbose but I think that's survivable.

shekyan:
@domenic, thanks, LGTM in regards to the change to strictly splitting algorithm, and now I know how a PR to infra should look like!

annevk:
@shekyan could you push one more commit adding your name to the acknowledgments?

shekyan:
@annevk, updated acknowledgements.

annevk:
Could you please keep it in alphabetical order? Thanks!

shekyan:
@annevk, updated it, sorry.

annevk:
Thank you! Alright, so the next step would be cleaning this up in HTML. Could maybe tack that onto https://github.com/whatwg/html/pull/2061 once I get motivated again.

PR #51, opened 2017-01-10
Link "string"s to their definitions

xfq:
Title says it all.

BTW IMHO the definition of string ("a string is a sequence of code points") is not sufficiently rigorous. Maybe we can learn from the [Character Model](https://www.w3.org/TR/charmod/#sec-Indexing) spec.


annevk:
What is ambiguous about the definition? Note though that we have #1 about indexing a string in different ways.

annevk:
This change seems fine by the way, if you're happy to discuss the definition further in #1 or a new issue, I'm happy to merge this.

xfq:
@annevk I'll comment in #1.

PR #52, opened 2017-01-10
Fix the definition of ASCII lower/upper alpha

### LINKED ISSUE #50: "An ASCII lower alpha is a code point in the ran..."
https://infra.spec.whatwg.org/commit-snapshots/208e4e04632d0c8514a8b5f26f99c8472d7e836d/#example-code-point-notation

> An ASCII lower alpha
> is a code point in the range U\+0041 to U\+005A, inclusive\.
> 
> An ASCII upper alpha
> is a code point in the range U\+0061 to U\+007A, inclusive\. 

* U+0041 is Latin Capital Letter A
* U+005A is Latin Capital Letter Z
* U+0061 is Latin Small Letter a
* U+007A is Latin Small Letter z

So the first range should be ASCII upper alpha; the second range should be ASCII lower alpha.

xfq:
Closes #50.

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -246,10 +246,10 @@ range U+0061 to U+0066, inclusive.
 <p>An <dfn export>ASCII hex digit</dfn> is an <a>ASCII upper hex digit</a> or
 <a>ASCII lower hex digit</a>.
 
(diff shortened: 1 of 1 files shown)


annevk:
Thanks!

PR #53, opened 2017-01-12
Add DFNs for list item & map entry, key, & value

tobie:
Use terminology consistently across the spec. Additionally, replace
incorrect reference to WebIDL record "entries" by "mappings".

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -404,17 +404,18 @@ ordered sets; implementations can optimize based on the fact that the order is n
 
 <h3 id=maps>Maps</h3>
 
-<p>A <dfn export lt="ordered map|map">ordered map</dfn>, or sometimes just "map", is a specification
-type consisting of a finite ordered sequence of key/value pairs, with no key appearing twice. Each
-key/value pair is called an <dfn for=map export>entry</dfn>.
+<p>A <dfn export lt="ordered map|map">ordered map</dfn>, or sometimes just "map", is a
+specification type consisting of a finite ordered sequence of
+<dfn for=map export>key</dfn>/<dfn for=map export>value</dfn> pairs, with no key appearing twice.
+Each key/value pair is called an <dfn for=map export>entry</dfn>.
 
 <p class=note>As with <a>ordered sets</a>, by default we assume that maps must also be ordered for
 interoperability among implementations.
 
 <p>A literal syntax can be used to express <a>ordered maps</a>, by surrounding the contents with «[
 ]» delimiters, denoting each <a for=map>entry</a> as |key| → |value|, and separating entries with
-a comma. An indexing syntax can be used to look up and set values by providing a key inside square
-brackets.
+a comma. An indexing syntax can be used to look up and set <a for=map>values</a> by providing a
+<a for=map>key</a> inside square brackets.
 
 <p class=example id=example-map-notation>Let |example| be the <a>ordered map</a> «[
 "<code>a</code>" → `<code>x</code>`, "<code>b</code>" → `<code>y</code>` ]». Then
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -296,30 +296,30 @@ for working with them, in order to create common ground.
 <h3 id=lists>Lists</h3>
 
 <p>A <dfn export>list</dfn> is a specification type consisting of a finite ordered sequence of
-items.
+<dfn for=list lt=item export>items</dfn>.
```
domenic: It appears we're not entirely consistent but I think this would be best as `for=list,stack,queue,set`. The usage sites appear a bit inconsistent too (e.g. sometimes in this spec we use `for=queue` for queues, sometimes we use `for=list`. But don't worry about that. I just think it'd be good to get the proper for="" values on the new exported terms in this PR.

infra.bs
```diff
@@ -378,40 +379,42 @@ but conventionally, the following operations are used to operate on it, instead
...
+type consisting of a finite ordered sequence of key/value pairs, with no key appearing twice. Each
+key/value pair is called an <dfn for=map export>entry</dfn>.
 
 <p class=note>As with <a>ordered sets</a>, by default we assume that maps must also be ordered for
 interoperability among implementations.
 
 <p>A literal syntax can be used to express <a>ordered maps</a>, by surrounding the contents with «[
-]» delimiters, denoting each key/value pair as |key| → |value|, and separating pairs with a comma.
-An indexing syntax can be used to look up and set values by providing a key inside square brackets.
+]» delimiters, denoting each <a for=map>entry</a> as |key| → |value|, and separating pairs with
```
domenic: separating entries?

infra.bs
```diff
@@ -450,16 +453,17 @@ of running <a for=map>get the keys</a> on the map.
...
 
 <hr>
 
 <p>An <a>ordered map</a> whose keys are all <a>strings</a> and whose values are all of the same Web
 IDL type |TValue| can be <dfn lt="convert ordered map to record">converted to the corresponding
 <a>record type</a></dfn> <code>record&lt;|TKey|, |TValue|></code> by first converting all its keys
-to the appropriate Web IDL <a>string type</a> |TKey|, and then creating corresponding record entries
-for each pair of converted key/original value. [[!WEBIDL]]
+to the appropriate Web IDL <a>string type</a> |TKey|, and then creating corresponding record
+<a>mappings</a> for each pair of converted key/original value. [[!WEBIDL]]
```
domenic: Let's add `for=record` to be precise, even though it's not necessary.

infra.bs
tobie: Yeah, it wasn't clear to me whether you wanted to create some sort of inheritance structure or "expose" the term on each one of the list types.

infra.bs
tobie: arg missed that.

tobie:
Totally fine waiting for @annevk's feedback.

infra.bs
```diff
@@ -296,30 +296,30 @@ for working with them, in order to create common ground.
 <h3 id=lists>Lists</h3>
 
 <p>A <dfn export>list</dfn> is a specification type consisting of a finite ordered sequence of
-items.
+<dfn for="list,stack,queue,set" lt=item export>items</dfn>.
 
 <p>For notational convenience, a literal syntax can be used to express <a>lists</a>, by surrounding
-the list contents by « » characters and separating list items with a comma. An indexing syntax can
-be used by providing a zero-based index into a list inside square brackets.
+the list contents by « » characters and separating list <a for=list>items</a> with a comma. An
```
annevk: I wonder if this is not better phrased as "surrounding its [items] by « » characters and separating them with a comma"

infra.bs
```diff
@@ -296,30 +296,30 @@ for working with them, in order to create common ground.
...
+add the given <a for=list>item</a> to the beginning of the list.
 
 <p class=note>The above definitions are modified when the <a>list</a> is an <a>ordered set</a>; see
 below for <a for=set lt=append>ordered set append</a> and
 <a for=set lt=prepend>ordered set prepend</a>.
 
-<p>To <dfn export for=list,set>remove</dfn> an item from a <a>list</a> is to remove all items
-from the list that match a given condition, or do nothing if none do.
+<p>To <dfn export for=list,set>remove</dfn> an <a for=list>item</a> from a <a>list</a> is to remove
+all items from the list that match a given condition, or do nothing if none do.
```
annevk: xref items

infra.bs
```diff
@@ -378,40 +379,42 @@ but conventionally, the following operations are used to operate on it, instead
...
+<a for=list>contains</a> the given <a for=set>item</a>, or to perform the normal <a>list</a>
 <a for=list>prepend</a> operation otherwise.
 
 
 <h3 id=maps>Maps</h3>
 
 <p>A <dfn export lt="ordered map|map">ordered map</dfn>, or sometimes just "map", is a specification
-type consisting of a finite ordered sequence of key/value pairs, with no key appearing twice.
+type consisting of a finite ordered sequence of key/value pairs, with no key appearing twice. Each
+key/value pair is called an <dfn for=map export>entry</dfn>.
```
annevk: Should we also be able to address key and value somehow?

infra.bs
tobie: Oh, it seemed the standard was not to link to the same dfn twice in the same paragraph.

infra.bs
tobie: Doesn't that create an ambiguity as to whether it is the items or the list that gets surrounded?

infra.bs
tobie: Added that. Not sure the dfns are in the most elegant place ever.

PR #68, opened 2017-02-22
Remove Web IDL conversions

### LINKED ISSUE #61: Move Web IDL conversions to Web IDL
Infra should not depend on Web IDL IMO; it should stick with things that are universally applicable to all specs, including non-Web IDL-based ones.

domenic:
Fixes #61. These will move to Web IDL itself, instead.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/remove-webidl-stuff/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/3af939f..remove-webidl-stuff:9cfc76e.html)


domenic:
Web IDL PR up at https://github.com/heycam/webidl/pull/317. I didn't actually copy over these conversion algorithms, but instead talked about how you can implicitly treat lists and ordered maps as sequences and records as necessary, which IMO is better.

PR #73, opened 2017-03-17
Define JavaScript string and scalar value string

### LINKED ISSUE #1: Rethink strings
I need to study the various dependencies of strings and figure out what we want to do. It seems there's a couple kind of strings that probably need to be distinguished and named somehow:

* JavaScript strings - each code point is in the range U+0000 to U+FFFF
* scalar value strings - each code point is a scalar value
* byte strings - each code point is in the range U+0000 to U+00FF
* ASCII strings - each code point is an ASCII code point
* strings - each code point is a code point (I don't think we really have this in the platform even though Encoding defines this kind of string; we have a variant of this where valid surrogate pairs are treated as their own code point)

annevk:
And also surrogate code point, code unit, and cast (for strings). Fixes
#1.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/string/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/a21296b...7d155b7.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -327,17 +327,22 @@ is immaterial. <a>Strings</a> are denoted by double quotes and monospace font.
 <p class=example id=example-string-notation>"<code>Hello, world!</code>" is a string.
 
 <p>To <dfn export for="JavaScript string">convert</dfn> a <a>JavaScript string</a> into a
-<a>scalar value string</a>, replace any <a>surrogate code points</a> with U+FFFD.
-<span class=note>Per definition these are isolated surrogates.</span>
+<a>scalar value string</a>, replace any <a>surrogates</a> with U+FFFD.
 <!-- Obviates need for https://heycam.github.io/webidl/#dfn-obtain-unicode -->
 
-<p class=note>In documents using the Infra Standard, a <a>scalar value string</a> can always be used
-as <a>JavaScript string</a> implicitly since it is a subset. The reverse is only possible if the
-<a>JavaScript string</a> is known to not contain <a>surrogates</a>. (An implementation likely has to
-perform explicit conversion, depending on how it actually ends up representing
-<a lt="JavaScript string">JavaScript</a> and <a>scalar value strings</a>. It is even fairly typical
-for implementations to have multiple implementations of just <a>JavaScript strings</a> for
-performance reasons and reducing memory usage.)
+<p class=note>The replaced surrogates are always isolated surrogates, since the process of
+interpreting the JavaScript string as containing <a>code points</a> will have converted surrogate
+pairs into single non-surrogate code points.)
+
+A <a>scalar value string</a> can always be used as <a>JavaScript string</a> implicitly since it is a
+subset. The reverse is only possible if the <a>JavaScript string</a> is known to not contain
+<a>surrogates</a>; otherwise a <a for="JavaScript string" lt=convert>conversion</a> must be
+performed.
+
+<p class=note>An implementation likely has to perform explicit conversion, depending on how it
+actually ends up representing <a lt="JavaScript string">JavaScript</a> and
+<a>scalar value strings</a>. It is even fairly typical for implementations to have multiple
+implementations of just <a>JavaScript strings</a> for performance and memory reasons.
 
 <hr>
 
--- a/infra.bs
+++ b/infra.bs
@@ -26,6 +26,7 @@ Boilerplate: omit conformance, omit feedback-header, omit idl-index
 <pre class="anchors">
 urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262; type: dfn
     text: List; url: sec-list-and-record-specification-type
+    text: The String Type; url: sec-ecmascript-language-types-string-type
 </pre>
 
 
@@ -252,8 +253,11 @@ in parentheses. [[!UNICODE]]
 
 <p>In certain contexts <a>code points</a> are prefixed with "0x" instead of "U+".
 
-<p>A <dfn export>scalar value</dfn> is a <a>code point</a> that is not in the range
-U+D800 to U+DFFF, inclusive.
+<p>A <dfn export>surrogate code point</dfn> is a <a>code point</a> that is in the range U+D800 to
+U+DFFF, inclusive.
+
+<p>A <dfn export>scalar value</dfn> is a <a>code point</a> that is not a
+<a>surrogate code point</a>.
 
 <p>An <dfn export>ASCII code point</dfn> is a <a>code point</a> in the range U+0000 to U+007F,
 inclusive.
@@ -294,10 +298,43 @@ inclusive.
 
 <h3 id=strings>Strings</h3>
 
-<p>A <dfn export>string</dfn> is a sequence of <a>code points</a>. Strings are denoted by double
-quotes and monospace font.
+<p>A <dfn export>JavaScript string</dfn> is a sequence of unsigned 16-bit integers, also known as
+<dfn export lt="code unit">code units</dfn>.
+
+<p class=note>This is different from how the Unicode Standard defines "code unit". In particular it
+refers exclusively to how the Unicode Standard defines it for Unicode 16-bit strings. [[UNICODE]]
+
+<p>A <a>JavaScript string</a> can also be interpreted as containing <a>code points</a>, per the
+conversion defined in <a>The String Type</a> section of the JavaScript specification. [[!ECMA-262]]
+
+<p class=note>This conversion process converts surrogate pairs into their corresponding
+<a>scalar value</a> and maps isolated surrogates to their corresponding <a>code point</a>, leaving
+them effectively as-is.
+
+<p>A <dfn export>scalar value string</dfn> is a sequence of <a>scalar values</a>.
+
+<p class=note>A <a>scalar value string</a> is useful for any kind of I/O or other kind of operation
+where <a>UTF-8 encode</a> comes into play.
+<!-- It's also useful if you can imagine the subsystem to be implemented in Rust -->
+
+<p><dfn export lt=string>String</dfn> can be used to refer to either a <a>JavaScript string</a> or
+<a>scalar value string</a>, when it is clear from the context which is meant or when the distinction
+is immaterial. <a>Strings</a> are denoted by double quotes and monospace font.
 
 <p class=example id=example-string-notation>"<code>Hello, world!</code>" is a string.
+
+<p>To <dfn export for="JavaScript string">convert</dfn> a <a>JavaScript string</a> into a
+<a>scalar value string</a>, replace any <a>surrogate code points</a> with U+FFFD.
+<span class=note>Per definition these are isolated surrogates.</span>
+<!-- Obviates need for https://heycam.github.io/webidl/#dfn-obtain-unicode -->
+
+<p class=note>A <a>scalar value string</a> can always be used as <a>JavaScript string</a> implicitly
+since it is a subset. The reverse is only possible if the <a>JavaScript string</a> is known to not
+contain <a>surrogate code points</a>. (An implementation likely has to perform explicit conversion,
+depending on how it actually ends up representing <a lt="JavaScript string">JavaScript</a> and
+<a>scalar value strings</a>. It is even fairly typical for implementations to have multiple
+implementations of just <a>JavaScript strings</a> for performance reasons and reducing memory
+usage.)
 
 <p>An <dfn export>ASCII string</dfn> is a <a>string</a> whose <a>code points</a> are all
 <a>ASCII code points</a>.
--- a/infra.bs
+++ b/infra.bs
@@ -253,11 +253,10 @@ in parentheses. [[!UNICODE]]
 
 <p>In certain contexts <a>code points</a> are prefixed with "0x" instead of "U+".
 
-<p>A <dfn export>surrogate code point</dfn> is a <a>code point</a> that is in the range U+D800 to
-U+DFFF, inclusive.
+<p>A <dfn export>surrogate</dfn> is a <a>code point</a> that is in the range U+D800 to U+DFFF,
+inclusive.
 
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -216,8 +217,11 @@ in parentheses. [[!UNICODE]]
```
domenic: Should we mention that these definitions come from Unicode? It seems like we should give them credit for this edifice in some way.

infra.bs
```diff
@@ -258,11 +262,31 @@ inclusive.
...
 
-<p>A <dfn export>string</dfn> is a sequence of <a>code points</a>. Strings are denoted by double
-quotes and monospace font.
+<p>A <dfn export>JavaScript string</dfn> is a sequence of unsigned 16-bit integers, also known as
+<dfn export lt="code unit">code units</dfn>. A <a>JavaScript string</a> can also be interpreted as
+containing <a>code points</a>, per the conversion defined in <a>The String Type</a> section of the
+JavaScript specification. [[!ECMA-262]]
+
+<p>A <dfn export>scalar value string</dfn> is a sequence of <a>scalar values</a>. These
+<a>scalar values</a> can also be addressed as <a>code points</a>.
```
domenic: I'm not sure this last sentence is useful. Since every scalar value is a code point it seems kind of redundant.

infra.bs
```diff
@@ -258,11 +262,31 @@ inclusive.
...
+
+<p class=note>A <a>scalar value string</a> is useful for any kind of I/O or other kind of operation
+where <a>UTF-8 encode</a> comes into play.
```
domenic: I think we need to expand this reasoning a bit more, given how confused people are about when to use USVString in Web IDL.

infra.bs
```diff
@@ -258,11 +262,31 @@ inclusive.
...
+<!-- It's also useful if you can imagine the subsystem to be implemented in Rust -->
+
+<p><dfn export lt=string>String</dfn> can be used to refer to either a <a>JavaScript string</a> or
+<a>scalar value string</a>, when it is clear from the context which is meant or when the distinction
+is immaterial. <a>Strings</a> are denoted by double quotes and monospace font.
 
 <p class=example id=example-string-notation>"<code>Hello, world!</code>" is a string.
 
+<p>To <dfn export for=string>cast</dfn> a <a>JavaScript string</a> into a
```
domenic: Is there a less generic word for this procedure? Maybe one people are already using? I guess not since Web IDL has a very long phrase for it.

Maybe "scalarize"?

infra.bs
```diff
@@ -258,11 +262,31 @@ inclusive.
...
+<a>scalar value string</a>, replace any <a>surrogate code points</a> with U+FFFD.
+<!-- Obviates need for https://heycam.github.io/webidl/#dfn-obtain-unicode -->
+
+<p class=note><a for=string>Casting</a> a <a>scalar value string</a> into a <a>JavaScript string</a>
+happens implicitly as desired as it is a lossless operation.
```
domenic: You can't use "casting", at least not with an `<a>`, for this direction. You defined it to only go in the other direction.

I'd phrase this as "can be done implicitly, since it's a lossless operation". The double-as is a bit awkward.

infra.bs
annevk: We acknowledge them for code point explicitly and everything else builds on top of that. I'll reference them again for code unit though as you suggested.

infra.bs
annevk: What else could we say? I think the confusion is mostly around folks not realizing that UTF-8 encode only handles scalar values.

infra.bs
annevk: What is wrong with a generic term when coupled with for=? We could also use "convert" if you don't like "cast". The for should probably be "JavaScript string" though I suppose.

annevk:
@aphillips could you please take a look as well? Thanks!

infra.bs
```diff
@@ -258,11 +262,31 @@ inclusive.
 
 <h3 id=strings>Strings</h3>
 
-<p>A <dfn export>string</dfn> is a sequence of <a>code points</a>. Strings are denoted by double
-quotes and monospace font.
+<p>A <dfn export>JavaScript string</dfn> is a sequence of unsigned 16-bit integers, also known as
+<dfn export lt="code unit">code units</dfn>. A <a>JavaScript string</a> can also be interpreted as
```
annevk: Yes, note that it isn't even like UTF-16 due to lone surrogates.

infra.bs
domenic: I guess as long as the call sites always say "to a scalar value string" it works. Whereas if it was "scalarize" the call sites could do "scalarize _s_". Both are fine.

infra.bs
aphillips: It isn't UTF-16. It's what Unicode (TUS Section 2.7) calls a _Unicode 16-bit string_. It's worth glancing at that section of Unicode

infra.bs
```diff
@@ -258,11 +262,36 @@ inclusive.
...
+containing <a>code points</a>, per the conversion defined in <a>The String Type</a> section of the
+JavaScript specification. [[!ECMA-262]]
+
+<p class=note>This is different from how the Unicode specification defines "code unit". In
+particular it refers exclusively to an unsigned 16-bit integer as is similar to how "code unit" is
+defined for UTF-16. However, therefore it can also refer to a surrogate that is not part of a
+surrogate pair, which is not the case for "code unit" as defined for UTF-16. [[UNICODE]]
```
aphillips: I think this note may be factually incorrect. It is the case that UTF-16 does not allow isolated surrogates. But the definition of `code unit` allows values in the surrogate range (how could it not?). The UTF-16 encoding form doesn't permit them to be isolated if the text is to be well-formed. But _Unicode 16-bit strings_ are not required to be well-formed (see my comment above)

infra.bs
```diff
@@ -258,11 +262,36 @@ inclusive.
...
+<!-- It's also useful if you can imagine the subsystem to be implemented in Rust -->
+
+<p><dfn export lt=string>String</dfn> can be used to refer to either a <a>JavaScript string</a> or
+<a>scalar value string</a>, when it is clear from the context which is meant or when the distinction
+is immaterial. <a>Strings</a> are denoted by double quotes and monospace font.
 
 <p class=example id=example-string-notation>"<code>Hello, world!</code>" is a string.
 
+<p>To <dfn export for="JavaScript string">convert</dfn> a <a>JavaScript string</a> into a
+<a>scalar value string</a>, replace any <a>surrogate code points</a> with U+FFFD.
```
aphillips: I would probably add the word "unpaired" outside the anchor tag for clarity. I know that this phrasing is correct, but you never quite say that properly paired surrogate code units make a supplementary code point. I think it might be useful to say that this conversion results in a (well-formed) UTF-16 string.

infra.bs
```diff
@@ -258,11 +262,36 @@ inclusive.
...
+<!-- Obviates need for https://heycam.github.io/webidl/#dfn-obtain-unicode -->
+
+<p class=note>A <a>scalar value string</a> can always be used as <a>JavaScript string</a> implicitly
+since it is a subset. The reverse is only possible if the <a>JavaScript string</a> is known to not
+contain <a>surrogate code points</a>.
```
aphillips: Again, I think `surrogate code points` is misleading (but not factually incorrect) here.

I guess the name "scalar value string" is still bothering me. I tend to expect a scalar value string to be a sequence of code points--Unicode scalar values--but this is a UTF-16 string. I'm concerned that the implicit conversion here is just a semantic nicety, not something someone would actually implement.

annevk: No, once you interpret a JavaScript string as containing code points, it's not at all related to UTF-16 anymore. It's just a sequence of code points with paired surrogates replaced.

infra.bs
annevk: That's defined by JavaScript. I guess I could add a note saying what JavaScript does there though.

infra.bs
annevk: Thanks, I'll study that section.

annevk:
@aphillips I guess maybe you're asking that we explicitly point out that a scalar value string can be represented in UTF-8 (and JavaScript string cannot always) and that any implicit conversion in standards might actually have to be an explicit conversion in implementations? I guess that's probably reasonable to do, since it's a rather complicated topic. I'll see about adding some more notes.

I already fixed the incorrect bits around code units. I didn't realize / forgot Unicode defined Unicode 16-bit strings. Good to know!

annevk:
@aphillips could you please give this another pass? Would be much appreciated. I'd really like to get this right.

infra.bs
```diff
@@ -258,11 +262,44 @@ inclusive.
...
+
+<p class=note>This is different from how the Unicode Standard defines "code unit". In particular it
+refers exclusively to how the Unicode Standard defines it for Unicode 16-bit strings. [[UNICODE]]
+
+<p>A <a>JavaScript string</a> can also be interpreted as containing <a>code points</a>, per the
+conversion defined in <a>The String Type</a> section of the JavaScript specification. [[!ECMA-262]]
+
+<p class=note>This conversion process converts surrogate pairs into their corresponding
+<a>scalar value</a> and maps isolated surrogates to their corresponding <a>code point</a>, leaving
+them effectively as-is.
```
domenic: I would be very interested in seeing an example of this conversion in action, that illustrates both of these points.

infra.bs
```diff
@@ -216,8 +217,11 @@ in parentheses. [[!UNICODE]]
 
 <p>In certain contexts <a>code points</a> are prefixed with "0x" instead of "U+".
 
-<p>A <dfn export>scalar value</dfn> is a <a>code point</a> that is not in the range
-U+D800 to U+DFFF, inclusive.
+<p>A <dfn export>surrogate code point</dfn> is a <a>code point</a> that is in the range U+D800 to
```
domenic: Maybe the term should be "unpaired surrogate" (or "isolated surrogate"?) instead of "surrogate code point"? Not sure I really understand this part though.

It's a bit strange that the word "code point" is included in this term but not in "scalar value".

infra.bs
```diff
@@ -258,11 +262,44 @@ inclusive.
...
+is immaterial. <a>Strings</a> are denoted by double quotes and monospace font.
 
 <p class=example id=example-string-notation>"<code>Hello, world!</code>" is a string.
 
+<p>To <dfn export for="JavaScript string">convert</dfn> a <a>JavaScript string</a> into a
+<a>scalar value string</a>, replace any <a>surrogate code points</a> with U+FFFD.
+<span class=note>Per definition these are isolated surrogates.</span>
+<!-- Obviates need for https://heycam.github.io/webidl/#dfn-obtain-unicode -->
+
+<p class=note>A <a>scalar value string</a> can always be used as <a>JavaScript string</a> implicitly
```
domenic: Maybe lead with "In specifications, " since a large part of this note is contrasting specifications and implementations.

infra.bs
```diff
@@ -258,11 +262,44 @@ inclusive.
...
+since it is a subset. The reverse is only possible if the <a>JavaScript string</a> is known to not
+contain <a>surrogate code points</a>. (An implementation likely has to perform explicit conversion,
+depending on how it actually ends up representing <a lt="JavaScript string">JavaScript</a> and
+<a>scalar value strings</a>. It is even fairly typical for implementations to have multiple
+implementations of just <a>JavaScript strings</a> for performance reasons and reducing memory
+usage.)
+
 <p>An <dfn export>ASCII string</dfn> is a <a>string</a> whose <a>code points</a> are all
```
domenic: I'd add a `<hr>` here

infra.bs
annevk: No, it should just be surrogate. You need more context to know whether it's isolated or paired. I guess I could drop "code point". I added it mainly to make it a bit more unique.

infra.bs
annevk: Wouldn't ECMAScript be a better place for that? I suppose I can add something here though.

infra.bs
domenic: Hmm, so then I don't quite understand why "Per definition these are isolated surrogates." later is true. What is special about the surrogates inside a JavaScript string that makes them always isolated?

infra.bs
domenic: It probably would, but I expect people to look to Infra for this sort of thing, and I think the fact that we use slightly more rigorous terminology will help make any example here clearer.

infra.bs
annevk: If you talk about code points in a JavaScript string, the code units that represent paired surrogates will have been replaced by a single scalar value.

annevk: I guess that argues for adding that example you asked for. It's not hard and I'll do it later today.

domenic: Oh! So there is an implicit "interpret as containing code points" step. Yeah, I totally missed that.

infra.bs
```diff
@@ -327,17 +327,22 @@ is immaterial. <a>Strings</a> are denoted by double quotes and monospace font.
...
-<a>JavaScript string</a> is known to not contain <a>surrogates</a>. (An implementation likely has to
-perform explicit conversion, depending on how it actually ends up representing
-<a lt="JavaScript string">JavaScript</a> and <a>scalar value strings</a>. It is even fairly typical
-for implementations to have multiple implementations of just <a>JavaScript strings</a> for
-performance reasons and reducing memory usage.)
+<p class=note>The replaced surrogates are always isolated surrogates, since the process of
+interpreting the JavaScript string as containing <a>code points</a> will have converted surrogate
+pairs into single non-surrogate code points.)
+
+A <a>scalar value string</a> can always be used as <a>JavaScript string</a> implicitly since it is a
```
annevk: Missing `<p>`. I'm surprised Bikeshed didn't complain to me about "surrogate code point" above.

annevk:
I emailed @aphillips for a final review. I can probably wait until Monday with landing to give him some more time and then start working on the other bits that need doing around strings.

annevk:
We should also add him to the acknowledgments section before landing.

PR #76, opened 2017-03-17
Define byte-case-insensitive

annevk:
***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/byte-case-insensitive/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/f4d2de2...30e52fd.html)


annevk:
Fetch uses byte-uppercase. Not sure about lowercase.

PR #78, opened 2017-03-17
Say a bit more about iteration

### LINKED ISSUE #66: While true / loop until break
See https://github.com/whatwg/fullscreen/pull/72/files#r101013520 for a need for this.

Do we have cases like this in other specs, and what do they say?

annevk:
Fixes #66.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/iteration/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/f4d2de2...7ae97d4.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -119,6 +119,29 @@ will make the caller automatically rethrow the given value, if any, and thereby 
 caller's algorithm. Using prose the caller has the ability to "catch" the exception and perform
 another action.
 
+<h4 id=algorithm-iteration>Iteration</h4>
+
+<p>There's a variety of ways to repeat a set of steps until a condition is reached. The Infra
+Standard is not (yet) exhaustive on this, but here are some examples:
+
+<ul>
+ <li><p>For each, as demonstrated later in this section.
+
+ <li>
+  <p>While
+
+  <div class=example id=example-while>
+   <ol>
+    <li>
+     <p>While true:
+     <ol>
+      <li>If |condition| is "<code>met</code>", then <a for=iteration>break</a>.
+      <li>&hellip;
+     </ol>
+   </ol>
+  </div>
+</ul>
+
 <p>An iteration's flow can be controlled via requirements to
 <dfn export for=iteration>continue</dfn> or <dfn export for=iteration>break</dfn>.
 <a for=iteration>Continue</a> will skip over any remaining steps in an iteration, proceeding to the
--- a/infra.bs
+++ b/infra.bs
@@ -121,26 +121,24 @@ another action.
 
 <h4 id=algorithm-iteration>Iteration</h4>
 
-<p>There's a variety of ways to repeat a set of steps until a condition is reached. The Infra
-Standard is not (yet) exhaustive on this, but here are some examples:
+<p>There's a variety of ways to repeat a set of steps until a condition is reached.
 
-<ul>
- <li><p>For each, as demonstrated later in this section.
+<p class=warning>The Infra Standard is not (yet) exhaustive on this, please file an issue if you
+need something.
 
- <li>
-  <p>While
+<p>For each, as defined <a for=list lt="for each">for lists</a> (and derivatives) and
+<a for=map lt="for each">maps</a>.
 
-  <div class=example id=example-while>
-   <ol>
-    <li>
-     <p>While true:
-     <ol>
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -119,6 +119,29 @@ will make the caller automatically rethrow the given value, if any, and thereby
 caller's algorithm. Using prose the caller has the ability to "catch" the exception and perform
 another action.
 
+<h4 id=algorithm-iteration>Iteration</h4>
+
+<p>There's a variety of ways to repeat a set of steps until a condition is reached. The Infra
+Standard is not (yet) exhaustive on this, but here are some examples:
+
+<ul>
+ <li><p>For each, as demonstrated later in this section.
```
mikewest: "iteration" is defined for `list` and `ordered map`, and both define their own "For each". Do you intend to unify those in some way?

infra.bs
```diff
@@ -119,6 +119,29 @@ will make the caller automatically rethrow the given value, if any, and thereby
...
+
+ <li>
+  <p>While
```
mikewest: Is this something you intend to `<dfn>`? Should we be linking `While` in specs?

infra.bs
annevk: Yeah, and folks use adhoc for each as well. And various other ways to get iteration I found, such as "Repeat" instead of "While true". So I was thinking we start listing them first and then slowly formalize them over time.

infra.bs
annevk: I'm not sure.

infra.bs
domenic: Editorially, you could insert links to for each with `, as defined <a lt="for each" for="list">for lists</a> and <a lt="for each" for="map">for maps</a>` if you wanted.

My perspective: it's somewhat important to define "for each" over data structures since it reinforces the fact that those data structures have an order. Other iterations are less important.

infra.bs
domenic: I think maybe the value is in telling people to use "while" instead of "repeat" or "perform the following steps repeatedly" or similar. Infra has a lot of terms that can feel silly to link to, and this falls pretty far on the silly end of the spectrum.

infra.bs
```diff
@@ -119,6 +119,29 @@ will make the caller automatically rethrow the given value, if any, and thereby
...
+
+  <div class=example id=example-while>
+   <ol>
+    <li>
+     <p>While true:
```
domenic: This is not a great example because it's better rewritten as `While |condition| is "met"`. And unlike the foreach example, it doesn't actually do anything. And it uses break/continue before they are introduced. Maybe just an example like "while condition is met, perform operation" would be good for introducing while. Then we could do a separate "while true" + break example after introducing break/continue.

infra.bs
domenic: Oh, and if we're doing to make it a definition, we need to actually define it. E.g. as https://en.wikipedia.org/wiki/While_loop does.

annevk:
It seems ECMAScript always uses "Repeat" by the way. Should we deviate from that?

infra.bs
```diff
@@ -119,6 +119,27 @@ will make the caller automatically rethrow the given value, if any, and thereby
 caller's algorithm. Using prose the caller has the ability to "catch" the exception and perform
 another action.
 
+<h4 id=algorithm-iteration>Iteration</h4>
+
+<p>There's a variety of ways to repeat a set of steps until a condition is reached.
+
+<p class=warning>The Infra Standard is not (yet) exhaustive on this, please file an issue if you
```
domenic: comma should be semicolon; see https://en.wikipedia.org/wiki/Comma_splice

infra.bs
```diff
@@ -119,6 +119,27 @@ will make the caller automatically rethrow the given value, if any, and thereby
...
+need something.
+
+<p>For each, as defined <a for=list lt="for each">for lists</a> (and derivatives) and
```
domenic: Hmm, these look weird as stand-alone paragraphs; I think it'd be best to go back to a bulleted list

infra.bs
domenic: The "warning" is so bright red and scary. Note might be better?

annevk:
I addressed the comments, btw.

Issue #13, opened 2016-11-07
Record-like data structure


annevk:
I'd like URL record, request, and response to just be some data structure so you can more easily address their members.

They're basically maps with fixed keys or what JavaScript calls records. The values are mostly mutable still, but thus far they don't have things similar to methods.

domenic:
My tentative comment on this idea is at https://github.com/whatwg/infra/issues/21#issuecomment-260380736

annevk:
So a request and a response become a very large tuple? I guess that's reasonable.

domenic:
Hmm yeah, the internal versions at least. It's interesting how Web IDL objects get a bunch of associated fields, but purely spec objects become tuples. I guess that's working as intended though.

tobie:
Why not use maps for this?

annevk:
@tobie we don't want mutations in size.

annevk:
A problem with using tuples though is that the order being defined can be a big problem for changing the API shape over time. If folks start passing a request around as a literal, then it would be very hard to add additional fields (except at the end, but that's not always the best place).

PR #79, opened 2017-03-17
Add tuples

### LINKED ISSUE #21: Tuples
We use tuples in a couple places and they're very much like immutable ordered sets. The syntax is typically (element1, element2).

annevk:
Fixes #21.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/tuples/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/f4d2de2...865ee66.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -610,7 +610,7 @@ subsequent prose.
 <h3 id=structs>Structs</h3>
 
 <p>A <dfn export>struct</dfn> is a specification type consisting of a finite set of
-<dfn for=struct,tuple,pair>items</dfn>, each of which has a unique and immutable
+<dfn for=struct,tuple,pair lt=item>items</dfn>, each of which has a unique and immutable
 <dfn export for=struct,tuple,pair>name</dfn>.
 
 <hr>
@@ -631,21 +631,21 @@ instance with the name given to the <a>tuple</a>.
  usage is then:</p>
 
  <ol>
-  <li>Let |statusInstance| be status (200, `<code>OK</code>`).
+  <li>Let |statusInstance| be the status (200, `<code>OK</code>`).
   <li>Set |statusInstance| to (301, `<code>FOO BAR</code>`).
   <li>If |statusInstance|'s code is 404, then &hellip;
  </ol>
 </div>
 
 <p class=note>It is intentional that not all <a>structs</a> are <a>tuples</a>. Documents using the
-Infra Standard might need the flexibility to add new <a for=struct>names</a> to their <a>struct</a>
-without breaking literal syntax used by their dependencies.
+Infra Standard might need the flexibility to add new <a for=struct>names</a> to their struct
+without breaking literal syntax used by their dependencies. In that case a tuple is not appropriate.
 
 <hr>
 
 <p><a>Tuples</a> with two <a for=tuple>items</a> are also known as <dfn export lt=pair>pairs</dfn>.
 For <a>pairs</a>, a slightly shorter literal syntax can be used, separating the two
-<a for=pair>items</a> with a / character. They are otherwise identical to <a>tuples</a>.
+<a for=pair>items</a> with a / character.
 
 <p class=example id=example-pair>Another way of expressing our |statusInstance| tuple above would be
 as 200/`<code>OK</code>`.
--- a/infra.bs
+++ b/infra.bs
@@ -446,7 +446,7 @@ for working with them, in order to create common ground.
 <h3 id=lists>Lists</h3>
 
 <p>A <dfn export>list</dfn> is a specification type consisting of a finite ordered sequence of
-<dfn for="list,stack,queue,set" lt=item export>items</dfn>.
+<dfn export for=list,stack,queue,set,tuple lt=item>items</dfn>.
 
 <p>For notational convenience, a literal syntax can be used to express <a>lists</a>, by surrounding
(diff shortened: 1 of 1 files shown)


annevk:
This isn't quite right.

infra.bs
```diff
@@ -547,6 +547,27 @@ ordered sets; implementations can optimize based on the fact that the order is n
...
+
+<p>Some lists are designated as <dfn export lt=tuple>tuples</dfn>. A tuple is a <a>list</a> with the
+additional semantic that its <a for=list>size</a> must not change and that its
+<a for=tuple>items</a> can be indexed by <dfn export for=tuple>name</dfn>.
+
+<p>For notational convenience, a literal syntax can be used to express <a>tuples</a>, by surrounding
+the tuple with ( ) characters and separating the <a for=tuple>items</a> by comma. The
+<a for=tuple>items</a> can be preceded by <a for=tuple>name</a> and : character if the
+<a for=tuple>names</a> are not clear from context. An indexing syntax can be used by providing a
+zero-based index and named index into a list inside square brackets.
```
mikewest: Is a 0-based index something we need to support for tuples? Seems like they'd always be named.

annevk: Yeah, I guess you're right.

Your pair example is also interesting. I wonder if we should have dedicated support for those. There's numerous places that use the / syntax that way.

annevk:
A separate type seems fine. I just found myself repeating a bunch of the list language so I used that as a shortcut. I didn't extend any of the list accessers apart from item with for=tuple though.

I like your idea of using |status|'s x instead.

And yes, I think we should have dedicated pairs with their own syntax. I'll have another go on Monday I suppose.

annevk:
Does having pairs mean we have to refactor maps somehow? It uses the pair language already...

infra.bs
```diff
@@ -547,6 +548,38 @@ ordered sets; implementations can optimize based on the fact that the order is n
 <a for=list>prepend</a> operation otherwise.
 
 
+<h3 id=tuples>Tuples</h3>
+
+<p>A <dfn export>tuple</dfn> is a specification type consisting of a finite ordered
+immutable-sequence of <dfn for=tuple,pair>items</dfn>, each of which has a
```
mikewest: What is "immutable" about the tuple? The list of names? The values associated with those names?

infra.bs
```diff
@@ -547,6 +548,38 @@ ordered sets; implementations can optimize based on the fact that the order is n
 <a for=list>prepend</a> operation otherwise.
 
 
+<h3 id=tuples>Tuples</h3>
+
+<p>A <dfn export>tuple</dfn> is a specification type consisting of a finite ordered
```
mikewest: Is "specification type" defined somewhere?

infra.bs
```diff
@@ -547,6 +548,38 @@ ordered sets; implementations can optimize based on the fact that the order is n
...
+immutable-sequence of <dfn for=tuple,pair>items</dfn>, each of which has a
+<dfn export for=tuple,pair>name</dfn>.
+
+<p>For notational convenience, a literal syntax can be used to express <a>tuples</a>, by surrounding
+the tuple with ( ) characters and separating the <a for=tuple>items</a> by comma. The
```
mikewest: It seems strange that we'd end up with "`a/b`" for a pair, but `(a, b, c)` for a triple+. I guess that's consistent with existing usage (at least in Fetch), but maybe it's worth changing to `(a, b)` for everything?

infra.bs
annevk: The sequence is. So you cannot change the size.

infra.bs
annevk: No, @domenic borrowed it from ECMAScript I think. We might want to formalize that too.

infra.bs
annevk: That seems okay to me too. I don't mind changing Fetch around too much. It just seemed that a/b might be nice to have.

domenic:
I changed a bunch of editorial things; let me know what you think.

I don't think maps should be based on pairs. The difference is subtle but tuples/pairs are for defining new novel specification types. So I don't think "pair" is a specification type that map would use. Rather, "status" is a specification type that fetch would define and use. There is no generic "pair" type.

Maybe we can phrase the intro better in that light. Hmm, not sure how though.

annevk:
Added another commit. I think this ends up fixing #21 and #81, but not sure about #13 due to ordering and syntax. (Immutable) records still seem better there.

domenic:
I think we should work to make this fix #13 before merging. I don't want two separate primitives.

Maybe we should flip things. We'd change the name to record, or maybe struct. We'd say that by default things are named, and that you can omit the names if there is a strong convention for the order that you plan to always follow.

annevk:
Would you be okay with having structs and ordered structs, and only the later get to have syntax (and be pairs)?

domenic:
That sounds pretty good. Editorially, I'd do that as a separate sentence under the same heading, instead of a separate heading, similar to how I changed pairs to be. That's because IMO it is good to intentionally blur the lines here and say they're all basically the same thing.

annevk:
Done. I removed the name syntax from tuples as well since we haven't actually needed that anywhere thus far.

domenic:
Made some small tweaks. LGTM if you like those, with an updated commit message of course.

annevk:
@mikewest you're still on board?

PR #89, opened 2017-03-22
Consistently format code points

annevk:
See https://github.com/whatwg/url/issues/47.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/code-point-formatting/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/30d99d6...780f545.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -183,16 +183,19 @@ out as being exclusive (with "but not both").
 <p>A <dfn export>byte</dfn> is a sequence of eight bits, represented as a double-digit hexadecimal
 number in the range 0x00 to 0xFF, inclusive.
 
-<p class=example id=example-byte-notation>0x49 (I) when <a>UTF-8 decoded</a> becomes U+0049 (I).
+<p>An <dfn export>ASCII byte</dfn> is a <a>byte</a> in the range 0x00 (NUL) to 0x7F (DEL),
+inclusive. As illustrated, an <a>ASCII byte</a> may be followed by the representation outlined in
+the <a href=https://tools.ietf.org/html/rfc20#section-2>Standard Code</a> section of
+<cite>ASCII format for Network Interchange</cite>, between parenthesis. [[!RFC20]]
 
-<p>An <dfn export>ASCII byte</dfn> is a <a>byte</a> in the range 0x00 to 0x7F, inclusive.
+<p class=example id=example-byte-notation>0x49 (I) when <a>UTF-8 decoded</a> becomes U+0049 (I).
 
 <h3 id=byte-sequences>Byte sequences</h3>
 
 <p>A <dfn export>byte sequence</dfn> is a sequence of <a>bytes</a>, represented as a space-separated
-sequence of bytes. Byte sequences with bytes in the range 0x20 to 0x7F, inclusive, can alternately
-be written as a string, but using backticks instead of quotation marks, to avoid confusion with an
-actual <a>string</a>.
+sequence of bytes. Byte sequences with bytes in the range 0x20 (SP) to 0x7E (~), inclusive, can
+alternately be written as a string, but using backticks instead of quotation marks, to avoid
+confusion with an actual <a>string</a>.
 
 <div class=example id=example-byte-sequence-notation>
  <p>0x48 0x49 can also be represented as `<code>HI</code>`.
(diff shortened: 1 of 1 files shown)


annevk:
It would be nice to be able to do the same thing for byte values, but there are no standardized names that I know of. Perhaps we should come up with our own convention?

infra.bs
```diff
@@ -221,47 +223,51 @@ in parentheses. [[!UNICODE]]
 <div class=example id=example-code-point-notation>
  <p>The <a>code point</a> rendered as 🤔 is represented as U+1F914.
 
- <p>When referring to that <a>code point</a>, we might instead say "U+1F914 THINKING FACE (🤔)",
- instead of just "U+1F914", to provide extra context.
+ <p>When referring to that <a>code point</a>, we might say "U+1F914 (🤔)", to provide extra context.
```
domenic: I think we should allow referring to it with both. Also this just modifies the example without modifying the text above which states how to represent it.

infra.bs
```diff
@@ -290,8 +296,8 @@ the <a>string</a> with their corresponding <a>code point</a> in <a>ASCII upper a
 
 <hr>
 
-<p>To <dfn export>strip newlines</dfn> from a <a>string</a>, remove any U+000A LINE FEED and
-U+000D CARRIAGE RETURN <a>code points</a> from the <a>string</a>.
+<p>To <dfn export>strip newlines</dfn> from a <a>string</a>, remove any U+000A LF and U+000D CR
```
domenic: Why is "LF" and "CR" acceptable here instead of "LINE FEED" and "CARRIAGE RETURN"?

The official standard is apparently http://www.unicode.org/charts/PDF/U0080.pdf which lists the name as "&lt;control&gt;" O_o. Now I'm just confused.

infra.bs
```diff
@@ -221,47 +223,51 @@ in parentheses. [[!UNICODE]]
...
 <p>An <dfn export>ASCII lower hex digit</dfn> is an <a>ASCII digit</a> or a <a>code point</a> in the
-range U+0061 to U+0066, inclusive.
+range U+0061 (a) to U+0066 (f), inclusive.
 
 <p>An <dfn export>ASCII hex digit</dfn> is an <a>ASCII upper hex digit</a> or
 <a>ASCII lower hex digit</a>.
 
-<p>An <dfn export>ASCII upper alpha</dfn> is a <a>code point</a> in the range U+0041 to U+005A,
-inclusive.
+<p>An <dfn export>ASCII upper alpha</dfn> is a <a>code point</a> in the range U+0041 (A) to
```
domenic: Why is this "A" instead of, per http://www.unicode.org/charts/PDF/U0000.pdf, "LATIN CAPITAL LETTER A"?

infra.bs
annevk: I wanted to require either one of them so we keep things brief and consistent. I guess I can be okay with also allowing the very long form though, as long as we don't use it.

infra.bs
annevk: Because it's shorter and just as clear.

infra.bs
annevk: I think you can use either per http://ftp.unicode.org/Public/UNIDATA/NamesList.txt. Also, I guess you meant to reference http://www.unicode.org/charts/PDF/U0000.pdf?

domenic: Ah yeah I misread as U+009A.

I'm not sure what the = sign means. For non-control characters it seems to lead to a very verbose description. But I guess we could say "one of the names in that list"?

infra.bs
domenic: Oh, I missed that this was inside the parenthesis, so we don't use the names there.

infra.bs
annevk: Yeah, maybe over time we can standardize on which one to pick. We could also introduce `<dfn>`s for the common code points, so you can do `<a>comma</a>` and such.

annevk:
I looked on Twitter for help for bytes and there's a reference for everything below 0x80, which we could allow: http://man7.org/linux/man-pages/man7/ascii.7.html (we could maybe inline this as a table). It probably doesn't make sense to require adding those in parenthesis though, but allowing it seems nice since sometimes it can really help a bit. And for 0x80 and above I guess it's best to have nothing. Documents could define their own names for those that make sense in the context of their format.

domenic:
So the remaining feedback before merging this is around the smiley face example.

For bytes it appears that man page is largely trying to say something about the ASCII Standard. Wikipedia links to http://worldpowersystems.com/archives/codes/X3.4-1963/page5.JPG as the original ASCII standard; interesting to see the differences. I was hoping we could refer to something for the byte names but maybe inlining the table is the way to go? Or just referencing the man page, I dunno.

annevk:
We could use https://tools.ietf.org/html/rfc20?

annevk:
So to be clear:

1. For code points: encourage either parenthesis (when it can be rendered) or name (when it cannot), but also allow both or neither. State this normatively, not just by example.
2. For bytes: allow the names from RFC 20 between parenthesis.

domenic:
Yep, that sounds exactly right. (Nice find with RFC 20.)

annevk:
Made those changes. I also changed the acceptable range for representing a byte sequence as a string, to omit 0x7F DEL which you cannot really represent.

domenic:
Made some small tweaks in a last pass. LGTM if you're happy with them.

PR #96, opened 2017-03-23
Define exists for lists and boundaries of list [] syntax

### LINKED ISSUE #90: Deal with list[n] access where n is negative or => size
We should probably say that it's not possible or actually define what it would return. If we return something it would have to be value like "none" or some such, that doesn't mean anything else and needs to be explicitly dealt with.

annevk:
Fixes #90.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/list-boundaries/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/139484a...bf9633d.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -541,7 +541,7 @@ all of its items.
 
 <p>A <a>list</a> <dfn export for=list,stack,queue,set lt=contain|exists>contains</dfn> an
 <a for=list>item</a> if it appears in the list. We can also denote this by saying that, for a
-<a>list</a> |list| and an <a for=list>item</a> |item|, "|list|[|item|] <a for=list>exists</a>".
+<a>list</a> |list| and an index |index|, "|list|[|index|] <a for=list>exists</a>".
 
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -538,8 +539,9 @@ all items from the list that match a given condition, or do nothing if none do.
 <p>To <dfn export for=list,stack,queue,set>empty</dfn> a <a>list</a> is to <a for=list>remove</a>
 all of its items.
 
-<p>A <a>list</a> <dfn export for=list,stack,queue,set lt=contain>contains</dfn> an
-<a for=list>item</a> if it appears in the list.
+<p>A <a>list</a> <dfn export for=list,stack,queue,set lt=contain|exists>contains</dfn> an
+<a for=list>item</a> if it appears in the list. We can also denote this by saying that, for a
+<a>list</a> |list| and an <a for=list>item</a> |item|, "|list|[|item|] <a for=list>exists</a>".
```
domenic: Not for an item, for an index.

annevk:
Pushed.

Issue #82, opened 2017-03-20
What was the motive in choosing algorithm description approach?


LeonidVasilyev:
Infra Standard defines [pseudocode and algorithm description approach](https://infra.spec.whatwg.org/#algorithms) that is different from what I usually find in [academic papers](https://www.cs.umd.edu/~samir/498/vitter.pdf) or classic books like [Introduction to Algorithms](https://mitpress.mit.edu/books/introduction-algorithms). Standard also notes that described algorithms aren't intended to be performant:

> Conformance requirements phrased as algorithms or specific steps may be implemented in any manner, so long as the end result is equivalent. (In particular, the algorithms are intended to be easy to follow, and not intended to be performant.)

What was the motive in choosing this specific style of algorithms description?

annevk:
I think in part because English prose is always the way it's been done, so formalizing around that seems less disruptive than inventing a syntax. But English prose also tends to be more readable and leaves more room for being creative in case you need to do something novel. Furthermore, if we had a syntax such as used in that paper, we'd then have to standardize that syntax fully and be constrained by it. Typically what I find is that such pseudocode is not actually defined in detail and therefore more ambiguous than English.

The reason why we don't go for performance is because that is really complicated. E.g., consider the JavaScript standard. If they had to concern themselves with performance they'd have to define JITs, a dozen string types, etc. For CSS you might end up having to define multiple selector matching engines and fiddle with them over time as the characteristics of the web change. It's much easier to describe input -> output and leave the bits between to be optimized as implementations see fit. That also allows completely novel implementation approaches such as Servo.

domenic:
@annevk, I think the last paragraph there could be morphed into a useful non-normative note in the spec. Less sure about the first one.

annevk:
@LeonidVasilyev are you okay with closing this? The standard now includes a note on performance. I'm not quite sure what to say about pseudocode other than using it meaning we'd have to define how that works (and how you extend it), etc.

LeonidVasilyev:
@annevk yes, thank you.

Issue #95, opened 2017-03-23
Add basic JavaScript types


annevk:
We should add undefined, null, and boolean (true, false). We haven't made much type-value distinction thus far so I'm not quite sure how to formulate this. Anyone ideas?

annevk:
Split from #87.

domenic:
I don't think we should have undefined in specs unless you are specifically dealing with JavaScript. And I think JavaScript should be a level above Infra.

I might just define the three values for now (inside the variables section?) without worrying about a type system.

annevk:
It seems like it should be in "primitive data types", but maybe its own section called "primitive values"? Where would we define Number/BigInt?

domenic:
Hmm yeah, primitive data types is probably better. Probably no need for a separate section. We'd definitely put number/bigint there.

PR #97, opened 2017-03-23
Add "let" and "set for variables

### LINKED ISSUE #88: Define initializing variables and setting them?
I.e., let and set. We could do that with `<dfn>`, but I don't think we want to require documents to link instances.

annevk:
Also add a note on performance for #82.

Fixes #88.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/variables/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/82d933c...66bf51a.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -131,6 +131,16 @@ as well as the single algorithm would. Therefore performance is best left as a f
 over.
 
 
+<h3 id=variables>Variables</h3>
+
+<p>A variable is declared with "let" and changed with "set".
+
+<p class=example id=example-variable>Let |list| be a new <a>list</a>.</p>
+
+<p>The scope of a variable is the step |s| it's declared in and any further and nested steps of the
+iteration |s| is in.
+
+
 <h3 id=algorithm-control-flow>Control flow</h3>
 
 <p>The control flow of algorithms is such that a requirement to "return" or "throw" terminates the
--- a/infra.bs
+++ b/infra.bs
@@ -137,8 +137,20 @@ over.
 
 <p class=example id=example-variable>Let |list| be a new <a>list</a>.</p>
 
-<p>The scope of a variable is the step |s| it's declared in and any further and nested steps of the
-iteration |s| is in.
+<div class=example id=example-variable-null>
+ <ol>
+  <li><p>Let |value| be null.
+
+  <li><p>If |input| is a <a>string</a>, then set |value| to |input|.
+
+  <li><p>Otherwise, set |value| to |input|, <a>UTF-8 decoded</a>.
+
+  <li><p>Assert: |value| is a <a>string</a>.
+ </ol>
+</div>
+
+<p>Variables must not be used before they are declared. Variables are
+<a href=https://en.wikipedia.org/wiki/Scope_(computer_science)#Block_scope>block scoped</a>.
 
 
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -122,6 +122,24 @@ leading spaces" or "return false") are to be interpreted with the meaning of the
 manner, so long as the end result is equivalent. (In particular, the algorithms are intended to be
 easy to follow, and not intended to be performant.)
 
+<p class=note>Performance is tricky to get correct as it is influenced by user perception, computer
+architectures, and different types of input that can change over time in how common they are. For
+instance, a JavaScript engine likely has many different code paths for what is standardized as a
+single algorithm to optimize for speed on memory consumption. Standardizing all those code paths
```
domenic: I'd say "...a single algorithm, in order to optimize..."

infra.bs
```diff
@@ -122,6 +122,24 @@ leading spaces" or "return false") are to be interpreted with the meaning of the
...
+
+
+<h3 id=variables>Variables</h3>
+
+<p>A variable is declared with "let" and changed with "set".
+
+<p class=example id=example-variable>Let |list| be a new <a>list</a>.</p>
+
+<p>The scope of a variable is the step |s| it's declared in and any further and nested steps of the
+iteration |s| is in.
```
domenic: This isn't specific to iterations; if blocks are also impacted. We might also want to state that you shouldn't use variables before declaring them... not sure what the best phrasing is here. Here's an attempt...

> Variables must not be used before they are declared. If variables are declared in a nested set of steps, they must not be used in steps that are in higher nesting levels.
>
> Note: this is essentially the same as "block scoping" in many programming languages.

infra.bs
```diff
@@ -122,6 +122,24 @@ leading spaces" or "return false") are to be interpreted with the meaning of the
 manner, so long as the end result is equivalent. (In particular, the algorithms are intended to be
 easy to follow, and not intended to be performant.)
 
+<p class=note>Performance is tricky to get correct as it is influenced by user perception, computer
```
tobie: Is this note in this PR on purpose or by accident?

annevk: See OP?

infra.bs
tobie: You want to add specific solutions as to how to treat common issues such as if..else statements.

@domenic, I've recently seen you use the pattern:

```
1.   Let |x| be an uninitialized variable.
1.   When condition, then set |x| to 1.
1.  Otherwise, set |x| to 2.
```

Maybe that should be used as an example.

And maybe also define `uninitialized` here.

annevk: I think using null there can still be fine. Especially if you assert at the end that it cannot be null. However, we haven't defined null yet, see #95.

tobie: I still think this particular pattern should be called out. It's contentious in PRs all the time.

infra.bs
tobie: Oops. Sorry.

annevk:
The performance note is now #101. Will rebase this after that lands.

I guess I'll just use null for the example @tobie wants and we can give it meaning later. Seems okay since documents using the Infra Standard already happily use it too.

annevk:
I thought we decided not to link most algorithm terms? Anyway, if you want to link Assert we should do that separately.

Issue #87, opened 2017-03-20
More data types


annevk:
Should we define numbers and there various notation schemes? (Mathematical operators?)

It might also make sense to define null as being roughly analogous to JavaScript's null and a good initial value for variables.

domenic:
We may want to wait to see how https://github.com/littledan/proposal-integer/issues/10 goes with regard to ES conventions, where this is more of an issue.

I agree something to make it clearer when e.g. integer division or floating-point addition is being used would be helpful. Not sure what a good solution looks like.

Null is probably harmless to add.

domenic:
On "null": in https://github.com/whatwg/html/pull/2421 I actually went with "Let _serialized_ be an uninitialized value" because I wanted it to be very clear that every branch of the algorithm was going to end up initializing it somehow. That felt clearer than "null", especially for the other direction where `null` is a valid JavaScript value. Not sure if we want to take that as precedent though.

annevk:
That kind of edge case we could also catch with an Assert perhaps (which we should also add).

annevk:
Also, when this happens, revisit whether we need to move anything from https://html.spec.whatwg.org/#numbers.

annevk:
https://github.com/tc39/ecma262/pull/1135 is pretty close to done it seems. Once that lands we can start aligning things.

littledan:
The BigInt PRs are all landed. In the final notation, subscripts are used on values, not operations. Values without subscripts are Numbers. I'm not sure exactly how this should apply to web specs. cc @caiolima @ms2ger

annevk:
An issue is that IDL talks about many distinct numeric types. In prose this mostly comes down to integers with range restrictions or float/double. Perhaps if we can make that more concrete we could have integers, with range restrictions at the IDL boundaries, and Numbers, with range restrictions at the IDL boundaries. And the IDL boundaries would also take care of conversion (e.g., some integers become a Number, others a BigNum).

The type in specification-prose could be inferred from the IDL, but we could copy the explicit notation in case there's room for ambiguity.

littledan:
Hmm, I'm not sure if we ended up with any notation for those sorts of restrictions that would be meaningful to copy here, unfortunately.

annevk:
That's fine, I was thinking of those more of IDL-level Asserts that specification algorithms would have to abide by. User agents could turn them into more dedicated types as they see fit.

msporny:
[...]

PR #98, opened 2017-03-23
Define how "Assert:" works

annevk:
See #87 for direct inspiration, though we’ve been using this here and
there for a while now.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/assertions/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/139484a...3b61f95.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -206,7 +206,11 @@ iteration.
 <h3 id=assertions>Assertions</h3>
 
 <p>To improve readability, it can sometimes help to add assertions to algorithms, stating
-invariants. To do this, write "<code>Assert: </code>", followed by a statement that must be true.
+invariants. To do this, write "Assert:", followed by a statement that must be true. If the statement
+ends up being false that indicates an issue with the document using the Infra Standard that should
+be reported and addressed.
+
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -203,6 +203,18 @@ iteration.
 </div>
 
 
+<h3 id=assertions>Assertions</h3>
+
+<p>To improve readability, it can sometimes help to add assertions to algorithms, stating
+invariants. To do this, write "<code>Assert: </code>", followed by a statement that must be true.
```
domenic: This should not be `<code>`; just normal font is fine.

annevk:
Shall I add that as a note?

infra.bs
annevk: Elsewhere we say that strings should use code.

annevk: But I guess we haven't done it for algorithm keywords per se.

domenic: Yeah this isn't a string in the sense of the data type.

domenic:
Not sure if it's a note or normative; I could go either way.

annevk:
Added a bit.

PR #104, opened 2017-03-27
Define character as alias for code point

### LINKED ISSUE #85: Define character as alias of code point or stop using it
We're currently using the term character to define syntax. We should probably stop doing that and use the syntax we outline for code points. Slightly weird to be informal here while we require much more of others.

See also #6 on the topic of whether or not to stop using character altogether as something that means code point all or some of the time (it seems somewhat silly to make it mean code point for something where Unicode says the code point is a non-character, but not out of the question).

annevk:
Fixes #85.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/character/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/f1be763...6637ff9.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -273,11 +273,11 @@ contains, in the range 0x61 (a) to 0x7A (z), inclusive, by 0x20.
 
 <h3 id=code-points>Code points</h3>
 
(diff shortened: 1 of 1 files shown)


annevk:
Step 3 of https://encoding.spec.whatwg.org/#euc-jp-decoder for instance.

Issue #108, opened 2017-03-28
Add permanently undefined characters (noncharacters)


annevk:
And see if that allows aligning some HTML and URL text to use a shared primitive and become more clear at the same time.

annevk:
(Per https://github.com/whatwg/html/pull/2471.)

domenic:
With the definition HTML uses, i.e.

> The control characters are those whose Unicode "General_Category" property has the value "Cc" in the Unicode UnicodeData.txt data file. [UNICODE]

? SGTM, assuming "Cc" stands for "control character". https://en.wikipedia.org/wiki/Control_character is also interesting.

annevk:
Yeah, although I think I'd list the explicit ranges instead as we do elsewhere. (We could maybe flip that around for some of the others too and have the explicit ranges in notes instead...)

domenic:
Hmm interesting. I generally like the idea of delegating to Unicode but for controls it probably doesn't matter as they really shouldn't add any more of those. For whitespace though I think they recently added a new whitespace, causing the ES spec to update in interesting ways. Food for thought.

annevk:
This was fixed by ad1b87aecce01759096fcdbf6acc2bd6096c3168 from #114.

PR #114, opened 2017-03-29
Annevk/noncharacter

annevk:
***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/noncharacter/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/c484a4f...60d532a.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -325,6 +325,12 @@ inclusive.
 
 <p>A <dfn export>scalar value</dfn> is a <a>code point</a> that is not a <a>surrogate</a>.
 
+<p>A <dfn export>noncharacter</dfn> is a <a>code point</a> that is in the range U+FDD0 to U+FDEF,
+inclusive, and U+FFFE, U+FFFF, U+1FFFE, U+1FFFF, U+2FFFE, U+2FFFF, U+3FFFE, U+3FFFF, U+4FFFE,
+U+4FFFF, U+5FFFE, U+5FFFF, U+6FFFE, U+6FFFF, U+7FFFE, U+7FFFF, U+8FFFE, U+8FFFF, U+9FFFE, U+9FFFF,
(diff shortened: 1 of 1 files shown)


annevk:
It's not entirely clear to me what the best place for noncharacter is. Before we continue with this PR I'd first like to work on corresponding PRs for URL and HTML to make sure this actually helps.

annevk:
Commit message note: this fixes #108.

annevk:
HTML could maybe use a definition of controls that excludes ASCII whitespace.

annevk:
I think it helps in terms of readability in both, but someone else should give it a look over as well.

infra.bs
```diff
@@ -325,6 +325,12 @@ inclusive.
 
 <p>A <dfn export>scalar value</dfn> is a <a>code point</a> that is not a <a>surrogate</a>.
 
+<p>A <dfn export>noncharacter</dfn> is a <a>code point</a> that is in the range U+FDD0 to U+FDEF,
+inclusive, or U+FFFE, U+FFFF, U+1FFFE, U+1FFFF, U+2FFFE, U+2FFFF, U+3FFFE, U+3FFFF, U+4FFFE,
+U+4FFFF, U+5FFFE, U+5FFFF, U+6FFFE, U+6FFFF, U+7FFFE, U+7FFFF, U+8FFFE, U+8FFFF, U+9FFFE, U+9FFFF,
+U+AFFFE, U+AFFFF, U+BFFFE, U+BFFFF, U+CFFFE, U+CFFFF, U+DFFFE, U+DFFFF, U+EFFFE, U+EFFFF, U+FFFFE,
+U+FFFFF, U+10FFFE, or U+10FFFF.
```
domenic: Can we explain why these characters are chosen? Apparently they are "permanently undefined" per Unicode?

Still not sure about whether we should be defining explicit ranges or referencing Unicode.

annevk: Opened #117 to sort this out, but I don't think that needs to block this, as we don't explain "surrogate" or "scalar value" in much detail either.

PR #115, opened 2017-03-29
Define fingerprinting vector

### LINKED ISSUE #20: Tracking vector tracking
I'm not sure where exactly we'd want to put this. Thoughts?

zcorpan:
Fixes #20.

---

This is blocked by https://github.com/tabatkins/bikeshed/issues/964

***
<a href="https://whatpr.org/infra/115.html" title="Last updated on Jan 15, 2020, 8:12 AM UTC (1c84122)">Preview</a> | <a href="https://whatpr.org/infra/115/fbe57e3...1c84122.html" title="Last updated on Jan 15, 2020, 8:12 AM UTC (1c84122)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -151,6 +151,7 @@ user has installed, information which can vary considerably from user to user.)
 <p tracking-vector>Features that are defined in documents using the Infra Standard that can be used
 as a <dfn export>tracking vector</dfn> are marked as this paragraph is.
 
+
 <p>Other features in the platform can be used for the same purpose, including, but not limited to:
 
 <ul>
--- a/infra.bs
+++ b/infra.bs
@@ -151,7 +151,6 @@ user has installed, information which can vary considerably from user to user.)
 <p tracking-vector>Features that are defined in documents using the Infra Standard that can be used
 as a <dfn export>tracking vector</dfn> are marked as this paragraph is.
 
-
 <p>Other features in the platform can be used for the same purpose, including, but not limited to:
 
 <ul>
--- a/infra.bs
+++ b/infra.bs
@@ -103,6 +103,65 @@ specifications using the Infra Standard, the user agent is generally the client 
 implements the specification.
 
 
+<h3 id=privacy>Privacy concerns</h3>
+
+<p>Some features that are defined in documents using the Infra Standard might trade user convenience
+for a measure of user privacy.
+
+<p>In general, due to the internet's architecture, a user can be distinguished from another by the
+user's IP address. IP addresses do not perfectly match to a user; as a user moves from device to
+device, or from network to network, their IP address will change; similarly, NAT routing, proxy
+servers, and shared computers enable packets that appear to all come from a single IP address to
+actually map to multiple users. Technologies such as onion routing can be used to further anonymize
+requests so that requests from a single user at one node on the internet appear to come from many
+disparate parts of the network. [[RFC791]]
+
+<p>However, the IP address used for a user's requests is not the only mechanism by which a user's
+requests could be related to each other. Cookies, for example, are designed specifically to enable
+this, and are the basis of most of the web's session features that enable you to log into a site
+with which you have an account. More generally, any kind of cache mechanism or shared state,
+including but not limited to HSTS, the HTTP cache, grouping of connections, storage APIs, can and
+ought to be expected to be abused. [[COOKIES]] [[RFC6797]] [[STORAGE]]
+
+<p>There are other mechanisms that are more subtle. Certain characteristics of a user's system can
+be used to distinguish groups of users from each other. By collecting enough such information, an
+individual user's browser's "digital fingerprint" can be computed, which can be better than an IP
+address in ascertaining which requests are from the same user.
+
+<p>Grouping requests in this manner, especially across multiple sites, can be used for malevolent
+purposes, e.g., governments combining information such as the person's home address (determined from
+the addresses they use when getting driving directions on one site) with their apparent political
+affiliations (determined by examining the forum sites that they participate in) to determine whether
+the person should be prevented from voting in an election.
+
+<p>Since the malevolent purposes can be remarkably evil, user agent implementors and specification
+authors are strongly encouraged to minimize leaking information that could be used to fingerprint or
+track a user.
+
+<p>Unfortunately, as the first paragraph in this section implies, sometimes there is great benefit
+to be derived from exposing APIs that can also be abused for fingerprinting and tracking purposes,
+so it's not as easy as blocking all possible leaks. For instance, the ability to log into a site to
+post under a specific identity requires that the user's requests be identifiable as all being from
+the same user, more or less by definition. More subtly, though, information such as how wide text
+is, which is necessary for many effects that involve drawing text onto a canvas (e.g., any effect
+that involves drawing a border around the text) also leaks information that can be used to group a
+user's requests. (In this case, by potentially exposing, via a brute force search, which fonts a
+user has installed, information which can vary considerably from user to user.)
+
+<p tracking-vector>Features that are defined in documents using the Infra Standard that can be used
+as a <dfn export>tracking vector</dfn> are marked as this paragraph is.
+
+<p>Other features in the platform can be used for the same purpose, including, but not limited to:
+
+<ul>
+ <li>The exact list of which features a user agents supports.
+ <li>The maximum allowed stack depth for recursion in script.
+ <li>Features that describe the user's environment.
+ <li>The user's time zone.
+ <li>HTTP request headers.
+</ul>
+
+
 <h2 id=algorithms>Algorithms</h2>
 
 <p>Algorithms, and requirements phrased in the imperative as part of algorithms (such as "strip any


infra.bs
```diff
@@ -84,6 +85,73 @@ either width or height is zero"), means an inclusive "or" (implying "or both"),
...
+limited to:
+
+<ul>
+ <li>The exact list of which features a user agents supports.
+
+ <li>The maximum allowed stack depth for recursion in script.
+
+ <li>Features that describe the user's environment.
+
+ <li>The user's time zone.
```
tobie: Kind of feel uneasy with what's a really partial list. Maybe grab some more examples from https://www.w3.org/TR/fingerprinting-guidance/ (or even link to it?)

annevk: It's called out as a partial list though and the intent of having this formalism is to establish the full list by getting all standards to point out their vectors.

(Also, the document you point to doesn't mention time zone and seems somewhat work in progress?)

tobie: Yeah--not sure what the state/fate of that doc is.

Time zone seems like an example of a bigger category (maybe user prefs? which would include things like do not track status, granted permissions, etc.).

Other things you might also mention are HW capabilities.

aphillips: An item that isn't mentioned in either (and which has the potential to be more identifying) is Accept-Language. Consider adding it to the list.

annevk: Time zone is kinda standalone, since I believe it's the main thing that leaks from JavaScript (other than performance characteristics and oh my SharedArrayBuffer). Accept-Language is a good one too.

xfq: A more general term, like "HTTP header fields", may be better. `Accept`/`Accept-Language`/`Connection`/`User-Agent`/`DNT` and many other headers can be used for fingerprinting.

annevk:
Hey, @othermaciej left some useful feedback in https://github.com/whatwg/html/issues/3054#issuecomment-353048033 that suggests we should maybe rename this to "tracking vector" or some such.

othermaciej:
I think it might be useful to define both "fingerprinting vector" and "tracking vector" as a superset of "fingerprinting vector". The current definition seems to be just about fingerprinting. But we need a way to tag features like LocalStorage and BroadcastChannel, and those are tracking but not fingerprinting.

annevk:
@othermaciej should those exist separately? And do we want separate icons for both? Or would annotating all with "tracking vector" and mention the specifics on a per feature basis be sufficient?

othermaciej:
@annevk I am not sure. Browsers may care about fingerprinting vectors and other tracking vectors at different levels. For Safari/WebKit, we care about both, but recognize that fully mitigating fingerprinting is a harder problem. Marking the two cases differently might help get better review from interesting parties.

Note: to be consistent, probably every stateful storage mechanism that exists in the web platform should be marked "tracking vector", including but not necessarily limited to: Session Storage, Local Storage, Cache API, HTTP cookies, IndexedDB, the HTTP cache, HSTS, HTTP ETags, SQL Storage, plugins (since many have their own storage mechanisms), CSS visited link history, Service Workers, Shared Workers.

(I realize not all of these are defined or even mentioned in WHATWG Living Standards or other standards that make use of Infra).

infra.bs
npdoty: @annevk yes, guidance on mitigating browser fingerprinting is a work in progress, and welcomes feedback, although the editor's draft would be more useful to review:
https://w3c.github.io/fingerprinting-guidance/
That draft does include a list of categories of information that can be used for browser fingerprinting, drawn from the Tor design document and slightly expanded upon, although it's not an explicit list of every such feature.

annevk:
I'd like to push this forward as I think this is rather relevant these days. To keep it simple I suggest we go with "tracking vector" exclusively, defining it such that it is inclusive of fingerprinting. If at some point we establish that we need more categories we can cross that bridge then.

annevk:
I'm fine with keeping the icon Simon created and using it for a broader purpose, but we could replace it if someone is able to create something suitable under CC0.

zcorpan:
I now think [the fingerprint image](https://github.com/whatwg/whatwg.org/pull/64/files?short_path=e713bf7#diff-e713bf741986d8c7a1df5f9792bf0cdf) I created looks kinda ugly, so I wouldn't mind a competing proposal. 🙂

annevk:
I'd be happy to replace it if someone were to offer something under an equivalent license. Otherwise though, I'd rather move ahead with this than wait for something better to appear.

infra.bs
```diff
@@ -84,6 +85,73 @@ either width or height is zero"), means an inclusive "or" (implying "or both"),
...
+individual user's browser's "digital fingerprint" can be computed, which can be as good, if not
+better, as an IP address in ascertaining which requests are from the same user.
+
+<p>Grouping requests in this manner, especially across multiple sites, can be used for both benign
+(and even arguably positive) purposes, as well as for malevolent purposes. An example of a
+reasonably benign purpose would be determining whether a particular person seems to prefer sites
+with dog illustrations as opposed to sites with cat illustrations (based on how often they visit
+the sites in question) and then automatically using the preferred illustrations on subsequent
+visits to participating sites. Malevolent purposes, however, could include governments combining
+information such as the person's home address (determined from the addresses they use when getting
```
othermaciej: I am not sure there's consensus that this is a benign purpose. This very specific dog vs cat case could be seen as benign. But more generally, showing users content based on revealed preference inspires thoughts of identifying more sensitive characteristics. That would not be benign, at least per the [WebKit Tracking Prevention Policy](https://webkit.org/tracking-prevention-policy/) and the [Mozilla Anti tracking policy](https://wiki.mozilla.org/Security/Anti_tracking_policy).

infra.bs
```diff
@@ -84,6 +85,73 @@ either width or height is zero"), means an inclusive "or" (implying "or both"),
...
+
+<p>However, the IP address used for a user's requests is not the only mechanism by which a user's
+requests could be related to each other. Cookies, for example, are designed specifically to enable
+this, and are the basis of most of the Web's session features that enable you to log into a site
+with which you have an account.
+
+<p>There are other mechanisms that are more subtle. Certain characteristics of a user's system can
+be used to distinguish groups of users from each other; by collecting enough such information, an
+individual user's browser's "digital fingerprint" can be computed, which can be as good, if not
+better, as an IP address in ascertaining which requests are from the same user.
```
othermaciej: This acknowledges fingerprinting but not "covert state" like HSTS supercookies or in general abuse of stateful mechanisms that are not meant to hold an identifier. (Technically, explicit state intended for storage is also a tracking vector by many  definitions of tracking).

infra.bs
```diff
@@ -84,6 +85,73 @@ either width or height is zero"), means an inclusive "or" (implying "or both"),
...
+identifiable as all being from the same user, more or less by definition. More subtly, though,
+information such as how wide text is, which is necessary for many effects that involve drawing
+text onto a canvas (e.g. any effect that involves drawing a border around the text) also leaks
+information that can be used to group a user's requests. (In this case, by potentially exposing,
+via a brute force search, which fonts a user has installed, information which can vary
+considerably from user to user.)
+
+<p class=fingerprinting-vector>Features that are defined in documents using the Infra Standard that
+can be <dfn export lt="fingerprinting vector">used to fingerprint the user</dfn> are marked as this
+paragraph is.
```
othermaciej: This should probably be generalized to "tracking vector" if that's our intent. And maybe more mention of stateful tracking mechanisms (or even navigational tracking) should be added.

infra.bs
annevk: I've tried to cover shared state and caches now.

infra.bs
annevk: I removed that whole bit.

infra.bs
```diff
@@ -103,6 +103,65 @@ specifications using the Infra Standard, the user agent is generally the client
...
+device, or from network to network, their IP address will change; similarly, NAT routing, proxy
+servers, and shared computers enable packets that appear to all come from a single IP address to
+actually map to multiple users. Technologies such as onion routing can be used to further anonymize
+requests so that requests from a single user at one node on the internet appear to come from many
+disparate parts of the network. [[RFC791]]
+
+<p>However, the IP address used for a user's requests is not the only mechanism by which a user's
+requests could be related to each other. Cookies, for example, are designed specifically to enable
+this, and are the basis of most of the web's session features that enable you to log into a site
+with which you have an account. More general, any kind of cache mechanism or shared state, including
```
jyasskin:
```suggestion
with which you have an account. More generally, any kind of cache mechanism or shared state, including
```

infra.bs
```diff
@@ -103,6 +103,65 @@ specifications using the Infra Standard, the user agent is generally the client
...
+but not limited to HSTS, the HTTP cache, grouping of connections, storage APIs, can and ought to be
+expected to be abused. [[COOKIES]] [[RFC6797]] [[STORAGE]]
+
+<p>There are other mechanisms that are more subtle. Certain characteristics of a user's system can
+be used to distinguish groups of users from each other; by collecting enough such information, an
```
jyasskin:
```suggestion
be used to distinguish groups of users from each other. By collecting enough such information, an
```

infra.bs
```diff
@@ -103,6 +103,65 @@ specifications using the Infra Standard, the user agent is generally the client
...
+individual user's browser's "digital fingerprint" can be computed, which can be as good, if not
```
jyasskin: Is "individual user's browser's" adding information over just "individual browser's" or "browser installation's"?

jyasskin:
```suggestion
individual user's browser's "digital fingerprint" can be computed, which can be better than an IP address
```

Fingerprints can stably identify an exact device, which is way better than an IP address.

infra.bs
```diff
@@ -103,6 +103,65 @@ specifications using the Infra Standard, the user agent is generally the client
...
+better, as an IP address in ascertaining which requests are from the same user.
+
+<p>Grouping requests in this manner, especially across multiple sites, can be used for malevolent
+purposes, e.g., governments combining information such as the person's home address (determined from
+the addresses they use when getting driving directions on one site) with their apparent political
+affiliations (determined by examining the forum sites that they participate in) to determine whether
+the person should be prevented from voting in an election.
+
+<p>Since the malevolent purposes can be remarkably evil, user agent implementors and specification
+authors are strongly encouraged to consider how to minimize leaking information that could be used
```
jyasskin: "Consider" is weasel-wordy.
```suggestion
authors are strongly encouraged to minimize the exposure of information that could be used
```

infra.bs
```diff
@@ -103,6 +103,65 @@ specifications using the Infra Standard, the user agent is generally the client
...
+purposes, so it's not as easy as blocking all possible leaks. For instance, the ability to log into
+a site to post under a specific identity requires that the user's requests be identifiable as all
+being from the same user, more or less by definition. More subtly, though, information such as how
+wide text is, which is necessary for many effects that involve drawing text onto a canvas (e.g.,
+any effect that involves drawing a border around the text) also leaks information that can be used
+to group a user's requests. (In this case, by potentially exposing, via a brute force search, which
+fonts a user has installed, information which can vary considerably from user to user.)
+
+<p class=tracking-vector>Features that are defined in documents using the Infra Standard that can be
+used as <dfn export>tracking vector</dfn> are marked as this paragraph is.
```
jyasskin:
```suggestion
used as a <dfn export>tracking vector</dfn> are marked as this paragraph is.
```

infra.bs
```diff
@@ -103,6 +103,65 @@ specifications using the Infra Standard, the user agent is generally the client
...
+the addresses they use when getting driving directions on one site) with their apparent political
+affiliations (determined by examining the forum sites that they participate in) to determine whether
+the person should be prevented from voting in an election.
+
+<p>Since the malevolent purposes can be remarkably evil, user agent implementors and specification
+authors are strongly encouraged to consider how to minimize leaking information that could be used
+to fingerprint or track a user.
+
+<p>Unfortunately, as the first paragraph in this section implies, sometimes there is great benefit
+to be derived from exposing the APIs that can also be abused for fingerprinting and tracking
```
jyasskin:
```suggestion
to be derived from exposing APIs that can also be abused for fingerprinting and tracking
```

jyasskin:
I've confirmed in a local build that the **tracking vector** paragraph gets a fingerprint icon, so no more need to wait for https://github.com/tabatkins/bikeshed/issues/964#issuecomment-548602441.

infra.bs
annevk: If a browser is used by multiple people, it is.

annevk:
Well, I'd like it to also link and use the remote icon and such.

jyasskin:
Yep, but Bikeshed's link will go to text that this PR is adding, so merging this shouldn't block on that.

annevk:
I suppose that's true, although the styling leaves to be desired. https://github.com/whatwg/whatwg.org/pull/64 fixes that, but does depend on the other changes.

Is there a reason to merge this early? Hopefully @tabatkins will soon be able to address the remaining issues.

annevk:
Tab has now done so and will add the link to Infra once this PR is merged. That seems fine as only HTML is impacted and HTML does not depend on Bikeshed. I've created https://github.com/tabatkins/bikeshed/pull/1578 so we can start using an external image here. I'll also check all the other PRs again to ensure everything is in order for the big switch.

annevk:
As there's always one more thing: https://github.com/tabatkins/bikeshed/pull/1586.

annevk:
Apologies for the commit title. Fortunately the description does mention "tracking vector" so `git log --grep` won't miss it.

I was too enthusiastic about finally being able to land this.

PR #118, opened 2017-03-30
Define insert for lists

### LINKED ISSUE #116: Insertion into lists
If I want to define step 6.1 of https://dom.spec.whatwg.org/#concept-node-insert as mutating an underlying list (children of a node is already stated to be a list as well) I need to be able to insert.

Is that acceptable? Or is that problematic for some reason?

Having this will help fix https://github.com/whatwg/dom/issues/174. The alternative I can think of is that I replace the entire list with a new list and just move the items across. Since nothing actually holds a pointer to the actual list (other than data structures DOM controls) that could probably work as well, but would be rather elaborate.

annevk:
Fixes #116.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/list-insert/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/4b5c3f3...2ff44e2.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -637,6 +637,9 @@ or do nothing if none do.
 below for <a for=set lt=append>ordered set append</a>, <a for=set>prepend</a>, and
 <a for=set>replace</a>.
 
+<p>To <dfn export for=list,set>insert</dfn> an <a for=list>item</a> into a <a>list</a> at index is
+to add the given item to the list at the given index.
+
 <p>To <dfn export for=list,set>remove</dfn> zero or more <a for=list>items</a> from a <a>list</a> is
 to remove all items from the list that match a given condition, or do nothing if none do.
 
--- a/infra.bs
+++ b/infra.bs
@@ -638,7 +638,8 @@ below for <a for=set lt=append>ordered set append</a>, <a for=set>prepend</a>, a
 <a for=set>replace</a>.
 
 <p>To <dfn export for=list,set>insert</dfn> an <a for=list>item</a> into a <a>list</a> at index is
-to add the given item to the list at the given index.
+to add the given item to the list between the given index &minus; 1 and the given index. If the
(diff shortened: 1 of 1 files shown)


annevk:
Thoughts I had while working on this:

* Should index be more formalized just like item is? Should you be able to get an item's corresponding index? (Might be useful in for each as well, to be able to get hold of the index.)
* We could define "insert before" for sets, but probably not needed for now.
* Should we be much more formal about the actual operations or is this good enough?

domenic:
This doesn't work for me; as defined it looks like it replaces the item at that index. That's why I think it needs to be phrased as "before index" or "after index".

> Should index be more formalized just like item is?

Yeah, probably.

> Should you be able to get an item's corresponding index? (Might be useful in for each as well, to be able to get hold of the index.)

That doesn't make sense for lists in general since the same value can appear at multiple indices.

> Should we be much more formal about the actual operations or is this good enough?

It might be worth a bigger bug to track, but I'm not sure what we could say that would still be concise.

infra.bs
```diff
@@ -637,6 +637,10 @@ or do nothing if none do.
 below for <a for=set lt=append>ordered set append</a>, <a for=set>prepend</a>, and
 <a for=set>replace</a>.
 
+<p>To <dfn export for=list,set>insert</dfn> an <a for=list>item</a> into a <a>list</a> at index is
```
domenic: s/at an index/before an index/

annevk:
Before I land this I should probably make sure this is actually enough for DOM. Feels a bit wrong otherwise.

annevk:
Created https://github.com/whatwg/dom/pull/435 for DOM. It looks good enough.

annevk:
This should have landed before we landed the DOM PR. Oh well. It'll sort itself out in a few days probably.

PR #121, opened 2017-04-03
Add syntax for lists of integers

### LINKED ISSUE #111: Repeat four times
URL's IPv4 serializer uses this phrase and then in the loop it asks whether it's the fourth time (and then doesn't prepend a dot). Should that be rewritten or formalized?

annevk:
Fixes #111.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/integer-list/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/ad1b87a...52717f5.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -630,13 +630,6 @@ out-of-bounds, except when used with <a for=list>exists</a>.
 "<code>b</code>", "<code>c</code>", "<code>a</code>" ». Then |example|[1] is the <a>string</a>
 "<code>b</code>".
 
-<p>Additionally, a literal syntax can be used to express a <a>list</a> of integers: the range
-<var>n</var> to <var>m</var>, inclusive, creates a new <a>list</a> whose <a for=list>items</a> are
-all integers in the range <var>n</var> to <var>m</var>, inclusive.
-
-<p class=example id=example-list-range><a for=list>For each</a> <var>n</var> in the range 1 to 4,
-inclusive, &hellip;
-
 <hr>
 
 <p>To <dfn export for=list>append</dfn> to a <a>list</a> that is not an <a>ordered set</a> is to
@@ -738,6 +731,16 @@ ordered sets; implementations can optimize based on the fact that the order is n
 <p>To <dfn export for=set>replace</dfn> within an <a>ordered set</a> is to replace the first item
 from the set that matches a given condition with the given <a for=set>item</a> and remove the other
 matching items, or do nothing if none match.
+
+<hr>
+
+<p><dfn export lt="the range">The range</dfn> <var>n</var> to <var>m</var>, inclusive, creates a new
+<a>ordered set</a> containing all of the integers from <var>n</var> up to and including <var>m</var>
+in consecutively increasing order, as long as <var>m</var> is greather than or equal to
+<var>n</var>.
+
+<p class=example id=example-the-range><a for=set>For each</a> <var>n</var> in <a>the range</a> 1 to
+4, inclusive, &hellip;
 
 
 <h3 id=maps>Maps</h3>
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -630,6 +630,13 @@ out-of-bounds, except when used with <a for=list>exists</a>.
 "<code>b</code>", "<code>c</code>", "<code>a</code>" ». Then |example|[1] is the <a>string</a>
 "<code>b</code>".
 
+<p>Additionally, a literal syntax can be used to express a <a>list</a> of integers: the range
+<var>n</var> to <var>m</var>, inclusive, creates a new <a>list</a> whose <a for=list>items</a> are
+all integers in the range <var>n</var> to <var>m</var>, inclusive.
```
zcorpan: `[1, 4, 4]` would satisfy this definition.

domenic:
In addition to the problem pointed out by @zcorpan:

- This is not a literal syntax in the traditional sense, just a term.
- I think this should be a dfn'ed term.

infra.bs
annevk: I don't understand. (If that's true all our other ranges are problematic too...)

zcorpan: It would work if do `s/in the range/represent the range/`, or so, but is a bit circular as a definition.

To clarify the problem: the number 1 is **in the range** 1 to 4, so a list consisting of a single item 1 has all its items in the range 1 to 4.

annevk: I guess I could make it "all the integers"

annevk: I agree about circular though, but the alternative is to define it in terms of some kind of loop, which also seems somewhat bad.

domenic: One alternative is to do what is common in math textbooks and say something like "the list « _n_, _n_ + 1, ..., _m_ - 1, m »". But that notation (the ... part in particular) is undefined, and has some magic where in fact the list could contain as few as one items despite looking like it contains at most four.

Maybe...

> creates a new ordered set containing all of the integers from _n_ up to and including _m_

which is somewhat saying the same thing in different words, but works OK.

Also the definition should state that it's "for _m_ &geq; _n_".

annevk: I like that, this also argues for moving it under sets.

zcorpan: "creates a new ordered set containing all of the integers from *n* up to and including *m* in consecutively increasing order."

annevk:
We don't use that symbol at the moment, or things like > and <. Not sure we should start now.

infra.bs
```diff
@@ -739,6 +732,16 @@ ordered sets; implementations can optimize based on the fact that the order is n
...
 
+<hr>
+
+<p><dfn export lt="the range">The range</dfn> <var>n</var> to <var>m</var>, inclusive, creates a new
+<a>ordered set</a> containing all of the integers from <var>n</var> up to and including <var>m</var>
+in consecutively increasing order, as long as <var>m</var> is greather than or equal to
+<var>n</var>.
+
+<p class=example id=example-the-range><a for=set>For each</a> <var>n</var> in <a>the range</a> 1 to
+4, inclusive, &hellip;
```
zcorpan: Is saying "inclusive" when using this term required? Or unnecessary?

annevk: If we can use this for all ranges we define anywhere we could maybe start leaving it out, but then phrases like "is an ASCII whitespace" become somewhat ill-defined. So I figured I'd stay consistent with those uses, but not actually start linking them.

zcorpan: OK, seems reasonable.

annevk:
Sigh, wrote list rather than set in the commit message. Close enough I guess.

PR #126, opened 2017-04-19
Make set replacement more limited in scope

annevk:
Per discussion in https://github.com/whatwg/dom/issues/442.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/set-replace/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/fa61fb2...6969703.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -728,9 +728,16 @@ ordered sets; implementations can optimize based on the fact that the order is n
 <a for=list>contains</a> the given <a for=set>item</a>, or to perform the normal <a>list</a>
 <a for=list>prepend</a> operation otherwise.
 
-<p>To <dfn export for=set>replace</dfn> within an <a>ordered set</a> is to replace the first item
-from the set that matches a given condition with the given <a for=set>item</a> and remove the other
-matching items, or do nothing if none match.
+<p>To <dfn export for=set>replace</dfn> within an <a>ordered set</a>, given <var>item</var>,
+<var>set</var>, and <var>replacement</var>, is to run these steps:
+
+<ol>
+ <li><p>If <var>set</var>[<var>replacement</var>] <a for=set>exists</a>, then <a>remove</a>
+ <var>item</var> from <var>set</var>.
+
+ <li><p>Otherwise, if <var>set</var>[<var>item</var>] <a for=set>exists</a>, replace <var>item</var>
+ within <var>set</var> with <var>replacement</var>.
+</ol>
 
 <hr>
 
@@ -869,6 +876,7 @@ as 200/`<code>OK</code>`.
 
 <p>Many thanks to
 Addison Phillips,
+Aryeh Gregor,
 Dominic Farolino,
 Jake Archibald,
 Jungkee Song,
--- a/infra.bs
+++ b/infra.bs
@@ -729,15 +729,12 @@ ordered sets; implementations can optimize based on the fact that the order is n
 <a for=list>prepend</a> operation otherwise.
 
 <p>To <dfn export for=set>replace</dfn> within an <a>ordered set</a>, given <var>item</var>,
-<var>set</var>, and <var>replacement</var>, is to run these steps:
+<var>set</var>, and <var>replacement</var> is to: if <var>set</var> <a for=set>contains</a>
+<var>item</var> or <var>replacement</var>, then replace the first instance of either with
+<var>replacement</var> and <a>remove</a> all other instances.
 
-<ol>
- <li><p>If <var>set</var>[<var>replacement</var>] <a for=set>exists</a>, then <a>remove</a>
- <var>item</var> from <var>set</var>.
-
- <li><p>Otherwise, if <var>set</var>[<var>item</var>] <a for=set>exists</a>, replace <var>item</var>
- within <var>set</var> with <var>replacement</var>.
-</ol>
+<p class=example id=example-set-replace>Replacing "a" with "c" within a <a>set</a> « "a", "b", "c" »
+gives « "c", "b" ». Within « "c", "b", "a" » it gives « "c", "b" » as well.
 
 <hr>
 
--- a/infra.bs
+++ b/infra.bs
@@ -728,13 +728,14 @@ ordered sets; implementations can optimize based on the fact that the order is n
 <a for=list>contains</a> the given <a for=set>item</a>, or to perform the normal <a>list</a>
 <a for=list>prepend</a> operation otherwise.
 
-<p>To <dfn export for=set>replace</dfn> within an <a>ordered set</a>, given <var>item</var>,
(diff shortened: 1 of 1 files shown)


ayg:
Why go with Safari's behavior and not Firefox's?  I don't have any objections, just wondering if there was a specific reason for it.

annevk:
It seemed more elegant to do either remove or replace than to do both. I also suspected it would be faster, but that probably varies depending on how you implement these exactly.

ayg:
It seems like if you keep the first of the two, which is what Firefox does, you're guaranteed to have to do a single pass -- keep the first you run into, and discard the second.  If you keep the string-to-find and discard the string-to-replace, then if you run into the string-to-replace first you'll replace it, then have to backtrack once you find the string-to-find.  If you don't mind doing two passes anyway, then there's no difference in implementation difficulty, but if you want to do one pass, Firefox seems better.

@domenic Do you prefer Safari's behavior?  Why?

infra.bs
```diff
@@ -728,9 +728,16 @@ ordered sets; implementations can optimize based on the fact that the order is n
...
 <a for=list>prepend</a> operation otherwise.
 
-<p>To <dfn export for=set>replace</dfn> within an <a>ordered set</a> is to replace the first item
-from the set that matches a given condition with the given <a for=set>item</a> and remove the other
-matching items, or do nothing if none match.
+<p>To <dfn export for=set>replace</dfn> within an <a>ordered set</a>, given <var>item</var>,
+<var>set</var>, and <var>replacement</var>, is to run these steps:
+
+<ol>
+ <li><p>If <var>set</var>[<var>replacement</var>] <a for=set>exists</a>, then <a>remove</a>
```
domenic: Indexing notation doesn't work like this; only numeric indices are allowed.

infra.bs
```diff
@@ -728,9 +728,16 @@ ordered sets; implementations can optimize based on the fact that the order is n
...
+ <var>item</var> from <var>set</var>.
```
domenic: I think this needs an example (whichever way we decide), and it should say "the result is X, *not* Y" to highlight that the other interpretation is wrong.

annevk:
@ayg how does the "first of the two" work in a genuine set implementation? A set can't contain duplicates.

wycats:
I don't remember the precise details of my original request, but the motivation was to make it easier for us to allow updating a declarative dynamic class that looks like: `class="small {{icon}} green"` so that the attribute matches what was typed.

For example, if `{{icon}}` started out as `bug` and became `note`, we'd want `class` to start out as `"small bug green"` and end up as `small note green`.

With the APIs available at the time of my request, you'd do:

```js
element.classList.add("small", "bug", "green");

// later

element.classList.remove("bug");
element.classList.add("note");

element.className; // "small green note"
```

The intent of `replace`, when I asked for it, was to allow:

```js
element.classList.add("small", "bug", "green");

// later

element.classList.replace("bug", "note");

element.className; // "small note green"
```

I wasn't thinking about the case of replacing a token with an existing token, and don't feel super-strongly about it.

That said, from the perspective of my original request, if somebody writes `class="small green {{icon}}"` and `icon` goes from `bug` to `small`, I think I'd expect the first one to win.

That's because I'd expect the results of a replacement to be equivalent to a "from-scratch" execution, where items earlier added in the set would win out over items added later in the set.

wycats:
That said, I also think it's a little weird that `classList` can't represent (with full fidelity) some things that can be represented in HTML (namely, repeated copies of the same token).

zcorpan:
https://twitter.com/domenic/status/855103294900981761 is @domenic's twitter poll, which has a number of comments also. 

> 22% (b, c)
> 34% (c, b)

domenic:
> then have to backtrack once you find the string-to-find

I'm not sure that's quite accurate, at least if your ordered set has O(1) membership test (like JS's ordered sets do). If that's the case you can basically execute the algorithm as-is in this PR currently, which could even require zero passes if removal is O(1) as well.

ayg:
@annevk "First of the two tokens that are about to become the same."  I.e., first of "a" and "c" if you're replacing "a" with "c".  I also agree with wycats that I'd expect the first to be kept because that's what would happen if you set the class attribute to "c b c" directly.  It seems more of the respondents to domenic's survey also would expect the first to be kept.

@domenic Gecko uses an array of pointers, so we don't have O(1) membership testing or O(1) removal.  If you have O(1) search, then I agree, I don't see why this PR would complicate the implementation at all.  (Even if removal isn't O(1), you'll have to remove something either way.)  Given how few items are likely to be in a DOMTokenList, though, a data structure that has O(1) search is probably overkill, and slower in any common case than an array.

At the end of the day, I don't care that much and would prefer to have some behavior specced in stone as of today or tomorrow or something, so go with whatever you want.

annevk:
I think we'd need to actually ask @cdumez why Safari has the set behavior it has and if they're willing to change it if we think the non-Safari behavior is better somehow. Otherwise we'll just have this conversation again the day after tomorrow.

@cdumez the question is what (a, b, c).replace(a, c) yields. In Safari it is (b, c). Some here (and on Twitter) argue it should be (c, b). Thoughts?

cdumez:
@annevk I actually voted for (c, b) on Twitter :) I did not realize that Safari was doing (b, c). I do not think there was any important reason for us to return (b, c) here. iirc, I merely implemented the spec literally although this could be an oversight on my part.

I do not have any issue updating WebKit to return (c, b). From what I gather, this is Firefox's behavior already? How about Chrome?

ayg:
Chrome and Edge don't support replace().

cdumez:
Ok then, no objection here. I can align WebKit with Firefox.

annevk:
> That said, I also think it's a little weird that classList can't represent (with full fidelity) some things that can be represented in HTML (namely, repeated copies of the same token).

To not leave this comment unaddressed: that is by design. We don't preserve whitespace and other things without meaning either.

annevk:
Updated the PR to reflect what Firefox does.

cdumez:
Were bugs filed against engines that do not match? I cannot find the WebKit bug.

cdumez:
Also, is there WPT test coverage?

cdumez:
Updating test coverage via https://github.com/w3c/web-platform-tests/pull/5725.

cdumez:
I filed the following WebKit bug: https://bugs.webkit.org/show_bug.cgi?id=171388

annevk:
@cdumez see https://github.com/whatwg/dom/pull/444#issuecomment-297280553. I haven't had time to investigate the compatibility issue, but that's unrelated to this change.

annevk:
And I don't think a bug was filed and that's my bad. I think I should start insisting on filing bugs before the PR lands.

PR #128, opened 2017-04-26
Meta: use allow-2119 to silence Bikeshed

annevk:
***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/2119/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/6ec9014...6197fcc.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -94,7 +94,7 @@ leading spaces" or "return false") are to be interpreted with the meaning of the
  <p>For example, were the spec to say:</p>
 
  <blockquote>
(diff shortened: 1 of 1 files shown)


annevk:
It usually catches legitimate failures so I'm rather thankful for this feature. I wish I could exclude more of a subtree, but as Tab pointed out that has risks in it as well.

foolip:
Ahem, I didn't really check what the warning was. If it's warning about normative words in examples, that's great actually.

PR #145, opened 2017-08-15
Define forgiving-base64

annevk:
For use by window.btoa()/window.atob() and data: URLs. Much of this text originated in the HTML Standard.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/base64/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/d28f5cd...8a6d030.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -899,8 +899,8 @@ certain inputs.
 <p>To <dfn export>forgiving-base64 decode</dfn> given a string <var>data</var>, run these steps:</p>
 
 <ol>
- <li><p>Let <var>position</var> be a pointer into <var>data</var>, initially pointing at the start
- of the string.
+ <li><p>Let <var>position</var> be a <a>position variable</a> for <var>data</var>, initially
+ pointing at the start of <var>data</var>.
 
  <li><p>Remove all <a>ASCII whitespace</a> from <var>data</var>.
 
@@ -1020,7 +1020,6 @@ certain inputs.
    order, and then empty <var>buffer</var>.
 
    <li><p>Advance <var>position</var> by 1.
-
   </ol>
 
  <li>
--- a/infra.bs
+++ b/infra.bs
@@ -894,7 +894,7 @@ certain inputs.
   <ul class="brief">
    <li>U+002B (+)
    <li>U+002F (/)
-   <li><span>ASCII alphanumeric</span>
+   <li><a>ASCII alphanumeric</a>
   </ul>
 
   <p>then return failure.
--- a/infra.bs
+++ b/infra.bs
@@ -16,20 +16,6 @@ Abstract: The Infra Standard aims to define the fundamental concepts upon which 
 Boilerplate: omit conformance, omit feedback-header, omit idl-index
 </pre>
 
-<style>
-#base64-table {
-  white-space: nowrap;
-  font-size: 0.6em;
-  column-width: 6em;
-  column-count: 5;
-  column-gap: 1em;
-}
-#base64-table thead { display: none; }
-#base64-table * { border: none; }
-#base64-table tbody td:first-child:after { content: ':'; }
-#base64-table tbody td:last-child { text-align: right; }
-</style>
-
 <script src=https://resources.whatwg.org/file-issue.js async></script>
 <script src=https://resources.whatwg.org/commit-snapshot-shortcut-key.js async></script>
 <script src=https://resources.whatwg.org/dfn.js defer></script>
@@ -899,13 +885,11 @@ certain inputs.
 <p>To <dfn export>forgiving-base64 decode</dfn> given a string <var>data</var>, run these steps:</p>
 
 <ol>
- <li><p>Let <var>position</var> be a <a>position variable</a> for <var>data</var>, initially
- pointing at the start of <var>data</var>.
-
  <li><p>Remove all <a>ASCII whitespace</a> from <var>data</var>.
+ <!-- https://lists.w3.org/Archives/Public/public-whatwg-archive/2011May/0207.html -->
 
  <li>
-  <p>If <var>data</var> contains a code point that is not one of
+  <p>If <var>data</var> contains a <a>code point</a> that is not one of
 
   <ul class="brief">
    <li>U+002B (+)
@@ -919,8 +903,8 @@ certain inputs.
   <p>If <var>data</var>'s <a for=string>length</a> divides by 4 leaving no remainder, then:
 
   <ol>
-   <li><p>If <var>data</var> ends with one or two U+003D (=) characters, then remove them from
-   <var>data</var>.
+   <li><p>If <var>data</var> ends with one or two U+003D (=) <a>code points</a>, then remove them
+   from <var>data</var>.
   </ol>
 
  <li><p>If <var>data</var>'s <a for=string>length</a> divides by 4 leaving a remainder of 1, then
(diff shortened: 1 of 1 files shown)


annevk:
@ayg this might be of interest since you researched the algorithm for `atob()` back in the day.

annevk:
By the way, I'd like to delay landing this until the tests are ready and the corresponding HTML PR is approved.

annevk:
I'm not sure why the build process started failing. Still fails after a restart.

infra.bs
```diff
@@ -872,6 +886,156 @@ For <a>pairs</a>, a slightly shorter literal syntax can be used, separating the
...
+  </ol>
+
+ <li><p>If <var>data</var>'s <a for=string>length</a> divides by 4 leaving a remainder of 1, then
+ return failure.
+
+ <li><p>Let <var>output</var> be an empty <a>byte sequence</a>.
+
+ <li><p>Let <var>buffer</var> be an empty buffer that can have bits appended to it.
+
+ <li>
```
domenic: Move the "Let _position_ be..." step to right before here, as otherwise it's a bit confusing what happens to it while the string mutates.

infra.bs
```diff
@@ -16,6 +16,20 @@ Abstract: The Infra Standard aims to define the fundamental concepts upon which
 Boilerplate: omit conformance, omit feedback-header, omit idl-index
 </pre>
 
+<style>
+#base64-table {
+  white-space: nowrap;
+  font-size: 0.6em;
+  column-width: 6em;
+  column-count: 5;
+  column-gap: 1em;
```
domenic: Sad that this doesn't work in Firefox, ugh. I tried Googling and messing with stuff but no luck.

infra.bs
```diff
@@ -872,6 +886,156 @@ For <a>pairs</a>, a slightly shorter literal syntax can be used, separating the
...
+
+ <li>
+  <p>While <var>position</var> does not point past the end of <var>data</var>:
+
+  <ol>
+   <li>
+    <p>Find the character pointed to by <var>position</var> in the first column of the following
+    table. Let <var>n</var> be the number given in the second cell of the same row.</p>
+
+    <div id="base64-table">
```
domenic: We could alternately say second column ... first cell of the same row in the RFC 4648 table 1.

infra.bs
```diff
@@ -872,6 +886,156 @@ For <a>pairs</a>, a slightly shorter literal syntax can be used, separating the
...
+<p>To <dfn export>forgiving-base64 decode</dfn> given a string <var>data</var>, run these steps:</p>
+
+<ol>
+ <li><p>Let <var>position</var> be a <a>position variable</a> for <var>data</var>, initially
+ pointing at the start of <var>data</var>.
+
+ <li><p>Remove all <a>ASCII whitespace</a> from <var>data</var>.
+
+ <li>
+  <p>If <var>data</var> contains a code point that is not one of
```
domenic: Link code point

infra.bs
```diff
@@ -872,6 +886,156 @@ For <a>pairs</a>, a slightly shorter literal syntax can be used, separating the
...
+   <li><span>ASCII alphanumeric</span>
+  </ul>
+
+  <p>then return failure.
+
+ <li>
+  <p>If <var>data</var>'s <a for=string>length</a> divides by 4 leaving no remainder, then:
+
+  <ol>
+   <li><p>If <var>data</var> ends with one or two U+003D (=) characters, then remove them from
```
domenic: s/characters/code points, with link

infra.bs
```diff
@@ -872,6 +886,156 @@ For <a>pairs</a>, a slightly shorter literal syntax can be used, separating the
...
+ <li><p>Let <var>output</var> be an empty <a>byte sequence</a>.
+
+ <li><p>Let <var>buffer</var> be an empty buffer that can have bits appended to it.
+
+ <li>
+  <p>While <var>position</var> does not point past the end of <var>data</var>:
+
+  <ol>
+   <li>
+    <p>Find the character pointed to by <var>position</var> in the first column of the following
```
domenic: s/character/code point, with link

infra.bs
```diff
@@ -872,6 +886,156 @@ For <a>pairs</a>, a slightly shorter literal syntax can be used, separating the
...
+    table. Let <var>n</var> be the number given in the second cell of the same row.</p>
+
+    <div id="base64-table">
+     <table>
+      <thead>
+       <tr>
+        <th>Character
```
domenic: s/Character/Code point, but no link in the header I think

infra.bs
```diff
@@ -872,6 +886,156 @@ For <a>pairs</a>, a slightly shorter literal syntax can be used, separating the
...
+
+<p>To <dfn export>forgiving-base64 encode</dfn> given a <a>byte sequence</a> <var>data</var>, apply
+the base64 algorithm defined in section 4 of RFC 4648 to <var>data</var> and return the result.
+[[!RFC4648]]
+
+<p class="note no-backref">This is named <a>forgiving-base64 encode</a> for symmetry with
+<a>forgiving-base64 decode</a>, which is different from the RFC as it defines error handling for
+certain inputs.
+
+<p>To <dfn export>forgiving-base64 decode</dfn> given a string <var>data</var>, run these steps:</p>
```
domenic: I think a note explaining that RFC 4648 actually doesn't contain a decode algorithm would be useful. (Certainly not one with error handling.) Cf. HTML's

```html
  <!-- Note: this function is defined explicitly here because RFC4648 does not specify how to handle
  erroneous input, and no preexisting browser implementation simply throws an exception on all
  erroneous input. -->
```

but I think having it as an actual note would be nice.

annevk: Isn't that what the note directly above it does?

infra.bs
annevk: Yeah, that might be better as long as we don't inline the encode algorithm.

infra.bs
domenic: I guess so, although by my reading of the RFC, it doesn't actually define any decode algorithm at all.

annevk: It defines an Encoding scheme and some rules around it. The idea is that you infer the decode and encode algorithms from that. Similar to using ABNF and expecting you have a parser that works.

domenic: I guess so. I'm OK with it as-is but I think it'd be nicer for our readers if you frame this as providing the missing decode algorithm. (There is a fairly explicit encode algorithm, in contrast.)

infra.bs
```diff
@@ -872,6 +872,83 @@ For <a>pairs</a>, a slightly shorter literal syntax can be used, separating the
...
+ <li><p>Remove all <a>ASCII whitespace</a> from <var>data</var>.
+ <!-- https://lists.w3.org/Archives/Public/public-whatwg-archive/2011May/0207.html -->
+
+ <li>
+  <p>If <var>data</var> contains a <a>code point</a> that is not one of
+
+  <ul class="brief">
+   <li>U+002B (+)
+   <li>U+002F (/)
+   <li><span>ASCII alphanumeric</span>
```
domenic: span should be a

infra.bs
```diff
@@ -872,6 +872,83 @@ For <a>pairs</a>, a slightly shorter literal syntax can be used, separating the
...
+   <li><p>If <var>data</var> ends with one or two U+003D (=) <a>code points</a>, then remove them
+   from <var>data</var>.
+  </ol>
+
+ <li><p>If <var>data</var>'s <a for=string>length</a> divides by 4 leaving a remainder of 1, then
+ return failure.
+
+ <li><p>Let <var>output</var> be an empty <a>byte sequence</a>.
+
+ <li><p>Let <var>buffer</var> be an empty buffer that can have bits appended to it.
```
domenic: Should we just make buffer a list? Then we can use append, and size (instead of "has accumulated"), and empty, and is empty.

annevk: But then the interpretation business gets a lot trickier. I'd rather leave this alone.

domenic: Eh, interpreting 24 bits as three 8-bit big-endian numbers seems to work fine whether those bits are in a buffer or in a list...

PR #146, opened 2017-08-15
byte sequence <> string

annevk:
This would improve https://github.com/whatwg/fetch/pull/579 and is also needed to properly define ByteString in IDL. There might be a few other places that can use it. This also makes polymorphic algorithms less needed as we just map back and forth.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/byte-sequence-string/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/d28f5cd...9570220.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -313,13 +313,10 @@ contains, in the range 0x61 (a) to 0x7A (z), inclusive, by 0x20.
 <a>byte sequence</a> <var>B</var>, if the <a>byte-lowercase</a> of <var>A</var> is the
 <a>byte-lowercase</a> of <var>B</var>.
 
-<p>To <dfn export for="byte sequence">map</dfn> a <a>byte sequence</a> <var>input</var> to a
-<a>string</a>, return a string whose <a for=string>length</a> is equal to <var>input</var>'s
+<p>To <dfn export>isomorphic decode</dfn> a <a>byte sequence</a> <var>input</var>, return a
+<a>string</a> whose <a for=string>length</a> is equal to <var>input</var>'s
 <a for="byte sequence">length</a> and whose code points have the same values as <var>input</var>'s
 bytes, in the same order.
-
-<p class=note>This matches the behavior of a ISO-8859-1 decoder, except no such decoder is defined
-as the Encoding Standard maps "<code>iso-8859-1</code>" to <a>windows-1252</a>. [[ENCODING]]
 
 <h3 id=code-points>Code points</h3>
 
@@ -458,8 +455,7 @@ implementations of just <a>JavaScript strings</a> for performance and memory rea
 
 <hr>
 
-<p>To <dfn export for=string>map</dfn> a <a>string</a> <var>input</var> to a <a>byte sequence</a>,
-run these steps:</p>
+<p>To <dfn export>isomorphic encode</dfn> a <a>string</a> <var>input</var>, run these steps:</p>
 
 <ol>
  <li><p>Assert: <var>input</var> contains no code points greater than U+00FF.
@@ -468,9 +464,6 @@ run these steps:</p>
  <var>input</var>'s <a for=string>length</a> and whose bytes have the same values as
  <var>input</var>'s code points, in the same order.
 </ol>
-
-<p class=note>This matches the behavior of a ISO-8859-1 encoder, except no such encoder is defined
-as the Encoding Standard maps "<code>iso-8859-1</code>" to <a>windows-1252</a>. [[ENCODING]]
 
 <hr>
 
--- a/infra.bs
+++ b/infra.bs
@@ -297,8 +297,9 @@ confusion with an actual <a>string</a>.
  <p>Headers, such as `<code>Content-Type</code>`, are <a>byte sequences</a>.
 </div>
 
-<p class=note>To get a <a>byte sequence</a> out of a <a>string</a>, use an operation such as
-<a>UTF-8 encode</a> from the Encoding Standard. [[ENCODING]]
+<p class=note>To get a <a>byte sequence</a> out of a <a>string</a>, using <a>UTF-8 encode</a> from
+the Encoding Standard is encouraged. In rare circumstances <a>isomorphic encode</a> might be needed.
+[[ENCODING]]
 
 <p>A <a>byte sequence</a>'s <dfn export for="byte sequence">length</dfn> is the number of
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -313,6 +313,14 @@ contains, in the range 0x61 (a) to 0x7A (z), inclusive, by 0x20.
 <a>byte sequence</a> <var>B</var>, if the <a>byte-lowercase</a> of <var>A</var> is the
 <a>byte-lowercase</a> of <var>B</var>.
 
+<p>To <dfn export for="byte sequence">map</dfn> a <a>byte sequence</a> <var>input</var> to a
```
domenic: I wonder about a name like "simple decode" or "numeric decode" or similar. "map" seems pretty generic and conflicts with my intuition that "the process of mapping a byte sequence to a string is called decoding".

infra.bs
```diff
@@ -313,6 +313,14 @@ contains, in the range 0x61 (a) to 0x7A (z), inclusive, by 0x20.
...
+<a>string</a>, return a string whose <a for=string>length</a> is equal to <var>input</var>'s
+<a for="byte sequence">length</a> and whose code points have the same values as <var>input</var>'s
+bytes, in the same order.
+
+<p class=note>This matches the behavior of a ISO-8859-1 decoder, except no such decoder is defined
+as the Encoding Standard maps "iso-8859-1" to <a>windows-1252</a>. [[ENCODING]]
```
domenic: We should be pretty cautious with this note as some people I know are very confused/mad by the mapping of iso-8859-1 to windows-1252. I'd rather phrase this in a way that makes it sound less like the Encoding Standard is being "wrong". And I think it's a valuable opportunity to spell out this confusing space for people. Including myself.

Something like

> This process corresponds somewhat to what has historically been called ISO-8859-1 decoding, at least for some ranges of bytes. However, in reality, software that performs such decoding has behavior for bytes in the range 128–256 that goes beyond the ISO-8859-1 specification, leading to the modern practice of aliasing "iso-8859-1" to "windows-1252" and performing the corresponding algorithm from the Encoding Standard. The algorithm here is different, treating all bytes directly as their corresponding code points instead of using the windows-1252 [index](https://encoding.spec.whatwg.org/#index) for bytes in the range 128–256. [[ENCODING]] [[ISO8859]].

([[ISO8859]] would go to https://www.iso.org/standard/28245.html I guess? Which doesn't let you view the PDF without paying? Wikipedia also gives ftp://std.dkuug.dk/JTC1/sc2/wg3/docs/n411.pdf as a link.)

Then we'll need a similar reverse note.

infra.bs
annevk: "Simple decode" seems bad because we should not use "simple". That's up to the reader to determine. And "numeric decode" also isn't really clear.

I'd be okay with "iso-8859-1 decode" and "iso-8859-1 encode". Those don't conflict with the Encoding Standard and a note can explain the difference with the iso-8859-1 label which maps to windows-1252.

infra.bs
```diff
@@ -313,6 +313,11 @@ contains, in the range 0x61 (a) to 0x7A (z), inclusive, by 0x20.
 <a>byte sequence</a> <var>B</var>, if the <a>byte-lowercase</a> of <var>A</var> is the
 <a>byte-lowercase</a> of <var>B</var>.
 
+<p>To <dfn export>isomorphic decode</dfn> a <a>byte sequence</a> <var>input</var>, return a
+<a>string</a> whose <a for=string>length</a> is equal to <var>input</var>'s
+<a for="byte sequence">length</a> and whose code points have the same values as <var>input</var>'s
```
domenic: Link code point

infra.bs
```diff
@@ -450,6 +455,18 @@ implementations of just <a>JavaScript strings</a> for performance and memory rea
 
 <hr>
 
+<p>To <dfn export>isomorphic encode</dfn> a <a>string</a> <var>input</var>, run these steps:</p>
+
+<ol>
+ <li><p>Assert: <var>input</var> contains no code points greater than U+00FF.
```
domenic: Link code point

infra.bs
```diff
@@ -450,6 +455,18 @@ implementations of just <a>JavaScript strings</a> for performance and memory rea
...
+
+ <li><p>Return a byte sequence whose <a for="byte sequence">length</a> is equal to
```
domenic: Link byte sequence

infra.bs
```diff
@@ -450,6 +455,18 @@ implementations of just <a>JavaScript strings</a> for performance and memory rea
...
+ <var>input</var>'s <a for=string>length</a> and whose bytes have the same values as
+ <var>input</var>'s code points, in the same order.
```
domenic: Link code point

infra.bs
```diff
@@ -450,6 +455,18 @@ implementations of just <a>JavaScript strings</a> for performance and memory rea
 
 <hr>
 
+<p>To <dfn export>isomorphic encode</dfn> a <a>string</a> <var>input</var>, run these steps:</p>
+
+<ol>
+ <li><p>Assert: <var>input</var> contains no code points greater than U+00FF.
+
+ <li><p>Return a byte sequence whose <a for="byte sequence">length</a> is equal to
+ <var>input</var>'s <a for=string>length</a> and whose bytes have the same values as
```
domenic: Link bytes

infra.bs
```diff
@@ -313,6 +313,11 @@ contains, in the range 0x61 (a) to 0x7A (z), inclusive, by 0x20.
 <a>byte sequence</a> <var>B</var>, if the <a>byte-lowercase</a> of <var>A</var> is the
 <a>byte-lowercase</a> of <var>B</var>.
 
+<p>To <dfn export>isomorphic decode</dfn> a <a>byte sequence</a> <var>input</var>, return a
+<a>string</a> whose <a for=string>length</a> is equal to <var>input</var>'s
+<a for="byte sequence">length</a> and whose code points have the same values as <var>input</var>'s
+bytes, in the same order.
```
domenic: Link bytes

annevk:
Since ISO-8859-1 is such a mess it seemed better to avoid getting into it. We can just teach people about isomorphic encode/decode and if they want to learn more that's more stackoverflow/IRC territory.

domenic:
I guess that's fair.

domenic:
Wait!

PR #148, opened 2017-08-30
Meta: update the URL for the common deploy.sh script

foolip:
Part of https://github.com/whatwg/whatwg.org/issues/71.

Also require the dist to be Trusty.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 language: generic
 
 env:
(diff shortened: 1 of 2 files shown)


annevk:
This one failed for some reason. Restarted.

annevk:
Still fails, not sure why.

foolip:
I also don't understand it. I think it's failing in the SERVICE_WORKER_SHA step, but it works locally. Maybe there's some rate limiting going on and this job is running on a machine that's being blocked?

foolip:
https://github.com/whatwg/compat/pull/84 is also failing in the same way. I'll add in some extra logging to debug in this PR.

foolip:
The problem was probably rate limiting, fixed in https://github.com/whatwg/whatwg.org/pull/74

PR #151, opened 2017-09-12
Editorial: add note about keyword suggestion for non-normative content

tobie:
This struck me as being useful to editors.

***
[Preview](https://s3.amazonaws.com/pr-preview/tobie/infra/informative.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/6509cfb...tobie:a4582f3.html)


annevk:
That's a great idea. Couple nits:

1. I'd make it the last paragraph of this section. As the current-last-paragraph stipulates the normative keywords we expect everyone to use.
2. I'd consider removing the note class.
3. I'd put the words in the same order as their normative counterparts. So from strong to weak.

PR #155, opened 2017-09-27
Add list clone, subsets, supersets, intersection and union

### LINKED ISSUE #154: Define superset, subset, union, intersection for ordered sets
Is this something you'd be interested to add, or are these obvious and don't need definitions?

tobie:
Closes #154.

***
[Preview](https://s3.amazonaws.com/pr-preview/tobie/infra/sets.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/c9cf341...tobie:c2dd0a1.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -715,6 +715,11 @@ performing a set of steps on each <a for=list>item</a> in order, use phrasing of
 "<a for=list>For each</a> |item| of <var ignore>list</var>", and then operate on |item| in the
 subsequent prose.
 
+<p>To <dfn export for=list,stack,queue,set>clone</dfn> a <a>list</a> |list| is to create a new
+<a>list</a> |clone| and, <a for=list>for each</a> |item| of |list|, <a for=list>append</a> |item|
+to |clone|, so that |clone| <a for=list>contains</a> the same <a for=list>items</a>, in the same
+order as |list|.
+
 <hr>
 
 <p>The <a>list</a> type originates from the JavaScript specification (where it is capitalized, as
--- a/infra.bs
+++ b/infra.bs
@@ -780,6 +780,12 @@ of either with <var>replacement</var> and <a for=set>remove</a> all other instan
 <a>ordered set</a> « "a", "b", "c" » gives « "c", "b" ». Within « "c", "b", "a" » it gives
 « "c", "b" » as well.
 
+<p>An <a>ordered set</a> |set| is a  <dfn export for=set>subset</dfn> of another <a>ordered set</a>
+|superset| (and conversely, |superset| is a <dfn export for=set>superset</dfn> of |set|)
+if, <a for=list>for each</a> |item| of |set|, |superset| <a for=set>contains</a> |item|.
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -715,6 +715,11 @@ performing a set of steps on each <a for=list>item</a> in order, use phrasing of
 "<a for=list>For each</a> |item| of <var ignore>list</var>", and then operate on |item| in the
 subsequent prose.
 
+<p>To <dfn export for=list,stack,queue,set>clone</dfn> a <a>list</a> |list| is to create a new
+<a>list</a> |clone| and, <a for=list>for each</a> |item| of |list|, <a for=list>append</a> |item|
+to |clone|, so that |clone| <a for=list>contains</a> the same <a for=list>items</a>, in the same
+order as |list|.
```
annevk: If you're making this explicit, you'd also have to explicitly clone the item. Or is the intent that they point to the same item?

infra.bs
```diff
@@ -775,6 +780,20 @@ of either with <var>replacement</var> and <a for=set>remove</a> all other instan
 <a>ordered set</a> « "a", "b", "c" » gives « "c", "b" ». Within « "c", "b", "a" » it gives
 « "c", "b" » as well.
 
+<p>An <a>ordered set</a> |set| is a <dfn export for=set>subset</dfn> of another <a>ordered set</a>
+|superset| (and conversely, |superset| is a <dfn export for=set>superset</dfn> of |set|) if,
+<a for=list>for each</a> |item| of |set|, |superset| <a for=set>contains</a> |item|. Note that this
+implies that an <a>ordered set</a> is both a <a for=set>subset</a> and a <a for=set>superset</a> of
+itself.
```
annevk: Note should be in a `<span class=note>` or its own paragraph.

infra.bs
tobie: That's always the question isn't it? I assumed we'd add deep cloning if the behavior you describe was ever required.

infra.bs
tobie: Will do.

infra.bs
annevk: I guess that's reasonable.

infra.bs
```diff
@@ -715,6 +715,11 @@ performing a set of steps on each <a for=list>item</a> in order, use phrasing of
 "<a for=list>For each</a> |item| of <var ignore>list</var>", and then operate on |item| in the
 subsequent prose.
 
+<p>To <dfn export for=list,stack,queue,set>clone</dfn> a <a>list</a> |list| is to create a new
```
domenic: I think we should expand on this in two ways:

> NOTE: this is a "shallow clone", as the items themselves are not cloned in any way

plus something about how cloning a specific type of list (stack/queue/ordered set) produces a list which is also one of those things. Maybe with an example?

tobie:
Do you want me to rebase those into 4 distinct commit (clone, subset & superset, intersection, union) or are you planning to mash them altogether into one (I don't care either way).

domenic:
I'd go for mash. Just wanting @annevk to sign off too before doing so.

PR #156, opened 2017-09-29
Define parse JSON from bytes

annevk:
Needed for Fetch, XMLHttpRequest, and probably Origin Policy, Manifest, and maybe more.

***
[Preview](https://infra.spec.whatwg.org/branch-snapshots/annevk/json/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/infra/c9cf341...c1d4b24.html)

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -90,7 +90,7 @@ readable.
 <p>In general, specifications interact with and rely on a wide variety of other specifications. In
 certain circumstances, unfortunately, conflicting needs require a specification to violate the
 requirements of other specifications. When this occurs, a document using the Infra Standard should
-denote such transgressions as a <dfn>willful violation</dfn>, and note the reason for that
+denote such transgressions as a <dfn export>willful violation</dfn>, and note the reason for that
 violation.
 
 <p class=example id=example-willful-violation>The previous section, [[#conformance]], documents a
@@ -897,13 +897,13 @@ as 200/`<code>OK</code>`.
 
 <h2 id=json>JSON</h2>
 
-<p>To <dfn>parse JSON from bytes</dfn> given <var>bytes</var>, run these steps:
+<p>To <dfn export>parse JSON from bytes</dfn> given <var>bytes</var>, run these steps:
 
 <ol>
  <li><p>Let <var>jsonText</var> be the result of running <a>UTF-8 decode</a> on <var>bytes</var>.
 
  <li>
-  <p>Return ? <a>Call</a>(<a>%JSONParse</a>, undefined, « <var>jsonText</var> »).
+  <p>Return ? <a>Call</a>(<a>%JSONParse%</a>, undefined, « <var>jsonText</var> »).
 
   <p class=note>The conventions used in this step are those of the JavaScript specification.
   [[ECMA-262]]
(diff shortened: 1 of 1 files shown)


annevk:
I wonder if it's worth pointing out that "web platform JSON" can start with a BOM.

cc @marcoscaceres @mikewest @dhausknecht @hsivonen

infra.bs
```diff
@@ -893,6 +895,21 @@ For <a>pairs</a>, a slightly shorter literal syntax can be used, separating the
 as 200/`<code>OK</code>`.
 
 
+<h2 id=json>JSON</h2>
+
+<p>To <dfn>parse JSON from bytes</dfn> given <var>bytes</var>, run these steps:
```
annevk: Add export=""

annevk:
Thankfully eating up the BOM is sanctioned per https://tools.ietf.org/html/rfc7159#section-8.1 even though it's not part of the grammar so I think we're good. And per http://software.hixie.ch/utilities/js/live-dom-viewer/saved/5399 it also matches implementations.

infra.bs
```diff
@@ -25,6 +25,8 @@ Boilerplate: omit conformance, omit feedback-header, omit idl-index
 
 <pre class="anchors">
 urlPrefix: https://tc39.github.io/ecma262/; spec: ECMA-262; type: dfn
+    text: %JSONParse%; url: sec-json.parse
+    text: Call; url: sec-call
```
domenic: `Call` should be `abstract-op`, not `dfn`

I guess we don't have a good name for stand-alone functions, so we can leave that as a dfn, despite that causing casing issues in "terms referenced by this specification". /cc @tabatkins

infra.bs
```diff
@@ -893,6 +895,21 @@ For <a>pairs</a>, a slightly shorter literal syntax can be used, separating the
...
 
+<h2 id=json>JSON</h2>
+
+<p>To <dfn export>parse JSON from bytes</dfn> given <var>bytes</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>jsonText</var> be the result of running <a>UTF-8 decode</a> on <var>bytes</var>.
+
+ <li>
+  <p>Return ? <a>Call</a>(<a>%JSONParse%</a>, undefined, « <var>jsonText</var> »).
```
domenic: Should be `<a abstract-op>` with the above changes, for Call.

domenic:
Hmm, in the branch snapshot [ENCODING] is still marked as an informative reference, despite you using UTF-8 decode in normative text... a Bikeshed issue, perhaps?

domenic:
Still LGTM but kind of annoying we have to add the explicit [specname]s. I guess I never was clear on the rules for those anyway.

annevk:
I think the references section is solely populated from `[[...]]` markup. I added one for Encoding.

annevk:
That's sort of the style that's been used forever, but I agree it's somewhat weird. We could consider alternatives, but probably best in a new issue.

tabatkins:
Hmm, there is indeed now one normative reference to Encoding in the spec, but it's still showing up in the Informative section (if I remove your manual normative reference). Some debugging statements shows that this is because it can't find the spec in the biblio db, but when I put the exact same autolink in a test document, it finds it just fine. This is... weird, to say the least. Further debugging required.

tabatkins:
Fixed now.

marcoscaceres:
File a bug to hook into this from manifest spec. Thanks for adding.

PR #171, opened 2017-11-23
Add null, true, and false

### LINKED ISSUE #95: Add basic JavaScript types

annevk:
Fixes #95.

***
<a href="https://pr-preview.s3.amazonaws.com/whatwg/infra/pull/171.html" title="Last updated on Nov 28, 2017, 5:11 PM GMT">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/whatwg/infra/aeadb8b...a1f4c51.html" title="Last updated on Nov 28, 2017, 5:11 PM GMT">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -275,6 +275,20 @@ be reported and addressed.
 
 <h2 id=primitive-data-types>Primitive data types</h2>
 
+<h3 id=nulls>Nulls</h3>
+
+<p>The value null is identical to the JavaScript <b>null</b> value. [[!ECMA-262]]
+
+<p class=example id=example-null>Let <var>element</var> be null.
+
+
+<h3 id=booleans>Booleans</h3>
+
+<p>A <dfn export>boolean</dfn> is either true or false.
+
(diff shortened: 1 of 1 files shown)


annevk:
(I'm not a 100% on the exact wording here, but I hope this will create some discussion. It's been a bit overdue.)

infra.bs
```diff
@@ -275,6 +275,20 @@ be reported and addressed.
 
 <h2 id=primitive-data-types>Primitive data types</h2>
 
+<h3 id=nulls>Nulls</h3>
+
+<p>The value null is identical to the JavaScript <b>null</b> value. [[!ECMA-262]]
```
domenic: I'd rather state what null is about. Maybe something like

> The value null is used to indicate "no value"

with an example that uses it more in context, e.g. return X if Y, otherwise return null.

We could state the JS connection as a separate sentence, like we do for Lists, but I think we should define null independently, instead of layering Infra on top of JS like this.

Issue #119, opened 2017-04-03
Labeled steps


annevk:
In https://github.com/whatwg/infra/issues/56#issuecomment-279854762 and further down the case is made that maybe labeled steps are useful.

The example given can be done with "break", but there might be more complicated examples with nested while loops that cannot.

(HTML also has some "goto label" in its algorithms, we should see about getting that refactored given http://www.u.arizona.edu/%7Erubinson/copyright_violations/Go_To_Considered_Harmful.html.)

PR #172, opened 2017-11-23
Add "Abort when" construct for Fetch and elsewhere

annevk:


annevk:
I suggest we first agree on this PR, then prepare a PR for Fetch (and maybe more) to see if it works in practice, and only when that's reviewed merge all of them.

domenic:
I felt inspired to do this with slightly different phrasing, and to add an example. Let me know what you think. The example relies on #119 for the desugaring, but I figure we'll be OK just using that without defining it for a little bit at least...

infra.bs
```diff
@@ -186,6 +186,79 @@ caller's algorithm. Using prose the caller has the ability to "catch" the except
...
 
 
+<h3 id=algorithm-conditional-abort>Conditional abort</h3>
+
+<p>Sometimes it is useful to stop performing a series of steps once some condition becomes true.
+
+<p>To do this, state that a given series of steps will <dfn export>abort when</dfn> a
+specific <var>condition</var> is reached. This indicates that the specified steps should be
+evaluated, not as-written, but by additionally inserting a step before each of them and after each
+of them except the last, that evaluates <var>condition</var>, and if <var>condition</var> evaluates
```
annevk: If we just said "before each of them" I think it ends up being the same and clearer?

annevk: Although I wonder if that's good enough. If the last step is computationally expensive it would be okay to abort during that step too...

infra.bs
```diff
@@ -186,6 +186,79 @@ caller's algorithm. Using prose the caller has the ability to "catch" the except
...
+  |result|.
+
+  <li><p><i>Done</i>: If |aborted| is true, then <a for=list>append</a> "<code>Didn't
+  finish!</code>" to |result|.
+ </ol>
+</div>
+
+<p class=note>Whenever this construct is used, implementations are allowed to evaluate
+<var>condition</var> during the specified steps rather than before and after each step, as long as
+the end result is indistinguishable from not doing that.
```
annevk: We should talk about the example here and mention that an implementation could for example abort during the compute operation, so long as it hasn't started mutating _result_.

annevk:
(I'll think a bit about doing this without labels. I'm still not convinced we should add them.)

infra.bs
domenic: Good point about before.

I think it's OK; if there's an abort before the step, and there are no observable side effects of the step, you can abort during the step. Allowing you to abort after the step doesn't give you any extra freedom, I don't think.

annevk:
(Found a way.)

Variable nits found by tooling are addressed by #195.

I'll see about putting together a PR for Fetch to accompany this.

annevk:
While going through Fetch I found:

> While true, breaking if the ongoing fetch terminates:

How does that translate?

jakearchibald:
I think that works fine.

1. Run these steps, but abort when condition:
    1. While true:
        1. Do some work.
1. If aborted, then…

annevk:
@jakearchibald ah yeah. Do you think "While true, but abort when condition" would be seen as confusing? Also, could you review this PR (or at least skim it)?

infra.bs
```diff
@@ -174,6 +174,78 @@ caller's algorithm. Using prose the caller has the ability to "catch" the except
 another action.
 
 
+<h3 id=algorithm-conditional-abort>Conditional abort</h3>
+
+<p>Sometimes it is useful to stop performing a series of steps once some condition becomes true.
```
jakearchibald: "a condition"?

infra.bs
```diff
@@ -174,6 +174,78 @@ caller's algorithm. Using prose the caller has the ability to "catch" the except
...
+
+<p>To do this, state that a given series of steps will <dfn export>abort when</dfn> a specific
+<var>condition</var> is reached. This indicates that the specified steps must be evaluated, not
+as-written, but by additionally inserting a step before each of them that evaluates
+<var>condition</var>, and if <var>condition</var> evaluates to true, skips the remaining steps.
```
jakearchibald: Do we need to make it clear that this is "deep"? As in, will be evaluated before each sub-step too.

infra.bs
```diff
@@ -174,6 +174,78 @@ caller's algorithm. Using prose the caller has the ability to "catch" the except
...
+
+<p>In such algorithms, the subsequent step can be annotated to run <dfn export>if aborted</dfn>, in
+which case it must run if the <var>condition</var> of the preceding <a>abort when</a> step evaluated
+to true.
```
jakearchibald: Might be clearer to say "if the preceding steps were skipped due to *condition* becoming true".

The ambiguity is if the preceding steps completed, but *condition* became true before the "if aborted" line.

infra.bs
```diff
@@ -174,6 +174,78 @@ caller's algorithm. Using prose the caller has the ability to "catch" the except
...
+   </ol>
+
+  <li><p>If the user clicked the "Cancel" button, then <a for=list>append</a>
+  "<code>Didn't finish!</code>" to |result|.
+ </ol>
+</div>
+
+<p class=note>Whenever this construct is used, implementations are allowed to evaluate
+<var>condition</var> during the specified steps rather than before and after each step, as long as
+the end result is indistinguishable from not doing that. For instance, as long as |result| in the
```
jakearchibald: Nit: Might not need "from not doing that"

infra.bs
```diff
@@ -174,6 +174,78 @@ caller's algorithm. Using prose the caller has the ability to "catch" the except
...
+
+<p>In such algorithms, the subsequent step can be annotated to run <dfn export>if aborted</dfn>, in
+which case it must run if the <var>condition</var> of the preceding <a>abort when</a> step evaluated
+to true.
+
+<div class=example id=example-conditional-abort>
+ <p>The following algorithm
+
+ <ol>
+  <li><p>Let |result| be an empty <a>list</a>.
```
jakearchibald: You're sometimes using `<var>whatever</var>`  and sometimes using `|whatever|`. Are you happy with doing that? 'Cos I'd love to use the shorthands in fetch spec stuff 😄.

annevk: I guess I'm okay with it. I'm not totally comfortable with embracing all of Bikeshed yet, but I guess this is fine.

infra.bs
annevk: Since we're trying to move away from substeps and haven't really discussed nested steps thus far I've not done this. I think that's okay for now, but welcome further discussion.

annevk:
@domenic can you sign off on this?

PR #177, opened 2017-11-28
Advise against using "for each" with stacks and queues

### LINKED ISSUE #149: Add a note that stacks are enumerated in FIFO order.
That's implicit because stacks are disguised lists, but it would be nice to clarify as that's arguably surprising.

domenic:
Closes #149.

***
<a href="https://pr-preview.s3.amazonaws.com/whatwg/infra/pull/177.html" title="Last updated on Nov 28, 2017, 5:40 PM GMT">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/whatwg/infra/ad786ae...d251a7d.html" title="Last updated on Nov 28, 2017, 5:40 PM GMT">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -754,6 +754,9 @@ but conventionally, the following operations are used to operate on it, instead 
 <a for=stack>item</a> and return it, if the <a>stack</a> <a for=stack>is not empty</a>, or to return
 nothing otherwise.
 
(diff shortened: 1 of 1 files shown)


annevk:
must not and remove them in the for= declaration? Or is there a reason it's still allowed?

domenic:
I didn't have a real reason why I did "should not", it just felt a little strong to prohibit it. But, better to be conservative, I suppose. Will fix.

PR #187, opened 2018-01-04
Define string concatenation

### LINKED ISSUE #185: String concenatation
How about:

> To string concatenate a list _list_ with an optional separator _separator_, run these steps:
> 1. If _list_ is empty, then return the empty string.
> 2. Assert: _list_ only contains strings.
> 3. If _separator_ is not given, then set _separator_ to the empty string.
> 4. Assert: _separator_ is a string.
> 5. Return a string whose contents are _list_'s items, in order, separated from each other by _separator_.

Needed for https://github.com/whatwg/dom/issues/555 and HTML; possibly other places. (URL concatenates bytes as well, but we don't have generics so...)

cc @Ms2ger

annevk:
Fixes #185.

***
<a href="https://whatpr.org/infra/187.html" title="Last updated on Jan 4, 2018, 6:48 PM GMT (ca8c862)">Preview</a> | <a href="https://whatpr.org/infra/187/4f17234...ca8c862.html" title="Last updated on Jan 4, 2018, 6:48 PM GMT (ca8c862)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -486,6 +486,24 @@ implementations of just <a>JavaScript strings</a> for performance and memory rea
 
 <hr>
 
+<p>To <dfn export for=string lt=concatenate|concatenation>concatenate</dfn> a <a for=/>list</a> of
+<a for=/>strings</a> <var>list</var>, with an optional separator string <var>separator</var>, run
+these steps:
+
+<ol>
+ <li><p>If <var>list</var> <a for=list>is empty</a>, then return the empty string.
+
+ <li><p>If <var>separator</var> is not given, then set <var>separator</var> to the empty string.
+
+ <li><p>Return a <a for=/>string</a> whose contents are <var>list</var>'s <a for=list>items</a>, in
+ order, separated from each other by <var>separator</var>.
+</ol>
+
+<p class=example id=example-string-concatenate>To serialize a set <var>set</var>, return the
+<a for=string>concatenation</a> of <var>set</var> with U+0020 SPACE.
+
+<hr>
+
 <p>To <dfn export>isomorphic encode</dfn> a <a>string</a> <var>input</var>, run these steps:</p>
 
 <ol>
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -486,6 +486,24 @@ implementations of just <a>JavaScript strings</a> for performance and memory rea
...
+ <li><p>If <var>list</var> <a for=list>is empty</a>, then return the empty string.
+
+ <li><p>If <var>separator</var> is not given, then set <var>separator</var> to the empty string.
+
+ <li><p>Return a <a for=/>string</a> whose contents are <var>list</var>'s <a for=list>items</a>, in
+ order, separated from each other by <var>separator</var>.
+</ol>
+
+<p class=example id=example-string-concatenate>To serialize a set <var>set</var>, return the
+<a for=string>concatenation</a> of <var>set</var> with U+0020 SPACE.
```
domenic: Here you are conflating a single code point and an entire string. Not sure whether we should be more careful, or just let it slide. (Maybe have a clause somewhere saying that single code points can be interpreted as strings when desired.)

infra.bs
```diff
@@ -486,6 +486,24 @@ implementations of just <a>JavaScript strings</a> for performance and memory rea
 
 <hr>
 
+<p>To <dfn export for=string lt=concatenate|concatenation>concatenate</dfn> a <a for=/>list</a> of
```
domenic: I envisioned this being located in the same hr-delimited section as the split operations.

infra.bs
annevk: Yeah, I was wondering about that. I was hoping we could just cast between those when needed. Same with sets and lists and such.

annevk: Happy to add text for this though.

domenic: Seeing the usage sites it seems like a good idea to add, perhaps in a separate PR.

infra.bs
```diff
@@ -486,6 +486,24 @@ implementations of just <a>JavaScript strings</a> for performance and memory rea
...
+<a for=/>strings</a> <var>list</var>, with an optional separator string <var>separator</var>, run
```
domenic: I wonder if "using" would be better than "with". "The concatenation of X with Y" almost sounds like the result should be "XY"

annevk: Good call.

infra.bs
annevk: Filed #188. Addressed the remainder.

Issue #198, opened 2018-05-07
Add a definition of multimap


BurtHarris:
The INFRA spec currently has a good definition of ordered map, but its definition requires key uniqueness.   Sometimes the uniqueness constraint isn't desirable.  For example whatwg/fetch describes [header-list](https://fetch.spec.whatwg.org/#concept-header-list) as multimap, saying:

> A header list is essentially a specialized multimap. An ordered list of key-value pairs with potentially duplicate keys. 

I would suggest it is useful to define **multimap** in the general whatwg/infra spec.   The language above, making multimap an ordered list of key-value pairs, seems perfectly acceptable to me.

annevk:
It would be interesting to see if a specialized data structure would simplify defining Headers/URLSearchParams/FormData a lot. Currently they use a list underneath, which is somewhat reasonable too.

domenic:
Headers went back to single-map, right?

annevk:
Ah, right, `Headers` did, but header list cannot (due to `Set-Cookie` which isn't exposed).

BurtHarris:
I drafted a pull request for this, but for some reason can't get bikeshed to process it (even the online version at https://api.csswg.org/bikeshed/).  The output page appears to contain the single character "#".   

I"m running Windows 10, so it might be a line ending problem, anyone familiar with this?

domenic:
I haven't had this problem myself; I also use Windows 10. But yes, please make sure you are using LF line endings; our gitattributes and editorconfig should be enforcing that.

Maybe first try compiling the spec as-is, to make sure that works?

BurtHarris:
@domenic, same result when I paste the original into https://api.csswg.org/bikeshed/ (as text.)   


If I use the upload file option, I get this message:


FATAL ERROR: Found unmatched text macro [SNAPSHOT-LINK]. Correct the macro, or escape it with a leading backslash.
 ✘ Did not generate, due to fatal errors

domenic:
Right, you need to set the SNAPSHOT-LINK macro, per the makefile.

BurtHarris:
Can you recommend what Make I should be using on Windows?

BurtHarris:
I seem to have a `curl` but it gives errors with the options used in the makefile.   Clearly some sort of toolset difference.

domenic:
You'll need https://curl.haxx.se/download.html. You can either copy and paste the commands from the Makefile, or use http://gnuwin32.sourceforge.net/packages/make.htm.

[...]

PR #199, opened 2018-05-07
Define multimap

BurtHarris:
Add term ordered multimap. 

Adjust language in definition of ordered map.

Fixes: #198

***
<a href="https://whatpr.org/infra/199.html" title="Last updated on May 7, 2018, 11:15 PM GMT (c0e7fcd)">Preview</a> | <a href="https://whatpr.org/infra/199/23ce6ea...c0e7fcd.html" title="Last updated on May 7, 2018, 11:15 PM GMT (c0e7fcd)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -922,13 +922,20 @@ in consecutively increasing order, as long as <var>m</var> is greather than or e
 
 <h3 id=maps>Maps</h3>
 
-<p>An <dfn export lt="ordered map|map">ordered map</dfn>, or sometimes just "map", is a
+<p>An <dfn export lt="ordered multimap|multimap">ordered multimap</dfn>, or sometimes just "multimap", is a
 specification type consisting of a finite ordered sequence of
-<dfn for=map export>key</dfn>/<dfn for=map export>value</dfn> pairs, with no key appearing twice.
+<dfn for=map export>key</dfn>/<dfn for=map export>value</dfn> pairs, with no requirement for key uniqueness, 
+nore that all equal keys appear together in the sequence.
 Each key/value pair is called an <dfn for=map export>entry</dfn>.
 
-<p class=note>As with <a>ordered sets</a>, by default we assume that maps need to be ordered for
-interoperability among implementations.
+<p>An <dfn export lt="ordered map|map">ordered map</dfn>, or sometimes just "map", is a
+specification type consisting of a finite ordered sequence of
+<a>key</a>/<a>value</a> pairs, with no key appearing twice.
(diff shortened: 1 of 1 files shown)


domenic:
Thanks, but without defining operations on these, this doesn't seem that useful.

Also in general I don't think we'd want to define something here unless it had a consumer, so as @annevk alluded to in #198, we'd need to see some patches against specs like XHR's FormData, Fetch's Headers, and URL's URLSearchParams to see if this could clean things up.

Probably surveying those specs would give you an idea of what operations would be useful to define for a multimap!

BurtHarris:
I see, but didn't we agree that HTML's header-list *was* a multimap?    
In HTML header-list (apparently) get's converted to a map using special rules if I understand things correctly.  Using semicolons as separators?  

See ietf/whatwg#25 for why I'm going here.   Trying to promote common terminology.

BurtHarris:
@domenic,  Fetch already refers to multimap, saying that **A header list is essentially a specialized multimap. An ordered list of key-value pairs with potentially duplicate keys.**   It just doesn't have a whatwg definition yet.

See cbor-wg/CBORbis#25 for why I'm proposing this here.   It would be so nice to share common terminology.

BurtHarris:
P.S.   it looks like https://www.npmjs.com/package/multimap has a start multimap operations, but I'm not sure I want to propose that without further research.  Defining the type's semantics seems more important.

domenic:
Fetch already has a WHATWG definition for multimap; Fetch is a WHATWG spec.

In general there's not much value in sharing a common definition for a data structure, unless that definition simplifies the specs that depend on it. The way we've done that for ordered map/set/list is by defining operations that can be reused among multiple instances. I think that is probably also possible for multimap. But this PR so far only duplicates the definition that already exists, giving us no gains. You haven't defined the type's semantics; you've just created a typedef against an existing type, which has no benefits over just using the existing type.

I would not suggest drawing multimap operations from an npm package. Instead, you should survey the specs I listed and see how they use multimaps. Are they performing operations in a verbose fashion, that would benefit from a simple definition? For example, most of https://fetch.spec.whatwg.org/#terminology-headers probably would apply to multimaps as well. The work in getting this PR accepted is pulling those in, then validating that the same thing applies to the other specs I pointed out, and that they would all benefit from shared definitions.

BurtHarris:
I see your point.    I'll have to take some time to incorporate this into my PR.

annevk:
Closing this as there hasn't been follow-up for a significant period of time. Happy to reconsider this though when you're ready.

PR #207, opened 2018-07-17
add 'JSON stringify and UTF-8 encode to bytes' algorithm

### LINKED ISSUE #193: Infra lacks a "JSON stringify and UTF-8 encode to bytes" operation
Infra features a [**parse JSON from bytes**](https://infra.spec.whatwg.org/#parse-json-from-bytes) operation, but lacks a "JSON stringify and UTF-8 encode to bytes" operation.

see also [webauthn](https://github.com/w3c/webauthn) issue [#712 JSON-serialized client data is wrong](https://github.com/w3c/webauthn/issues/712)

equalsJeffH:
this fixes #193.  

note that this attendant change to ECMA262 needs to also be accomplished: 
https://mail.mozilla.org/pipermail/es-discuss/2018-March/050324.html

see also: https://github.com/w3c/webauthn/issues/712

HTH, 

=JeffH

ps: we PayPal signed the whatwg participant agreement, see: https://github.com/whatwg/participant-data/blob/master/entities.json

***
<a href="https://whatpr.org/infra/207.html" title="Last updated on Aug 2, 2018, 9:03 PM GMT (5114d6e)">Preview</a> | <a href="https://whatpr.org/infra/207/49cfbc0...5114d6e.html" title="Last updated on Aug 2, 2018, 9:03 PM GMT (5114d6e)">Diff</a>


infra.bs
```diff
@@ -1042,6 +1042,19 @@ as 200/`<code>OK</code>`.
   <p class=note>The conventions used in this step are those of the JavaScript specification.
 </ol>
 
+<p>To <dfn export>JSON stringify and UTF-8 encode to bytes</dfn> a given JavaScript value
```
domenic: Maybe omit "and UTF-8 encode" for symmetry? Since we assume everything is UTF-8 these days? Or maybe we should keep it.

infra.bs
```diff
@@ -1042,6 +1042,19 @@ as 200/`<code>OK</code>`.
...
+<var>value</var>, run these steps:
+
+<ol>
+ <li>
+  <p>Let <var>jsonString</var> be the result of
+  <a abstract-op>Call</a>(<a>%JSONStringify%</a>, undefined, « <var>value</var> »).
+  [[!ECMA-262]]
+  <p class=note>The conventions used in this step are those of the JavaScript specification.
```
domenic: Maybe add `<p class=note>Since no additional arguments are passed to <a>%JSONStringify%</a>, the resulting string will have no whitespace inserted</p>`

infra.bs
```diff
@@ -1042,6 +1042,19 @@ as 200/`<code>OK</code>`.
   <p class=note>The conventions used in this step are those of the JavaScript specification.
 </ol>
 
+<p>To <dfn export>JSON stringify and UTF-8 encode to bytes</dfn> a given JavaScript value
+<var>value</var>, run these steps:
+
+<ol>
+ <li>
+  <p>Let <var>jsonString</var> be the result of
+  <a abstract-op>Call</a>(<a>%JSONStringify%</a>, undefined, « <var>value</var> »).
```
domenic: I think you'll need to add a link thingy to the top of the document for JSONStringify, like was done for JSONParse.

equalsJeffH:
my paypal2 org membership is already public AFAICT:
https://github.com/PayPal2
https://github.com/equalsJeffH

..?

infra.bs
equalsJeffH: I personally would prefer it to be explicit.  but y'all are editors of this spec, so up to you.

equalsJeffH: I personally would prefer it to be explicit.  but y'all are editors of this spec, so up to you. IIRC that is the phrase that Anne suggested a while back.

domenic:
Heh, I imagine you can see that, as you're a member. But if you use incognito mode or similar, you'll find that PayPal2 currently has no public members:

image


So, you need to follow the above link's instructions and publicize.

equalsJeffH:
> Heh, I imagine you can see that, as you're a member. But if you use incognito mode or similar, you'll find that PayPal2 currently has no public members.

doh, sorry. I think it is fixed now.

infra.bs
equalsJeffH: good catch, thx.

infra.bs
equalsJeffH: I added your suggested Note to the existing Note. if not OK will make it a separate note, see 7d94fda.

annevk:
We call the other "parse JSON from bytes". It would make sense then to start this one with "stringify". But I guess "stringify JSON to bytes" is a little weird.

I don't have a strong opinion here. It does seem that the downstream PR landed already, so if we make a change we'd have to change that too.

aphillips:
@annevk "stringify" and "to bytes" don't mentally go together to me, since to me the starting state is the string, while the bytes are just data (payload). How about "serialize"?

annevk:
Yeah that works, it's just not what JavaScript^WECMAScript calls it.

domenic:
I'm a fan of serialize. I'll update this, merge, and also submit a PR to webauth as recompense for us being wishy washy over here :)

equalsJeffH:
thanks guys :)

PR #209, opened 2018-07-27
Add ASCII encode and decode algorithms.

jyasskin:
This isn't that important, but in case you want it:

I'm using this with
https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-4.2,
which returns strings in some cases that I'd like to use them as byte
sequences. Because of
https://tools.ietf.org/html/draft-ietf-httpbis-header-structure-07#section-3.7,
I'm guaranteed they're actually all ASCII.

***
<a href="https://whatpr.org/infra/209.html" title="Last updated on Jul 30, 2018, 5:44 PM GMT (9232e78)">Preview</a> | <a href="https://whatpr.org/infra/209/49cfbc0...9232e78.html" title="Last updated on Jul 30, 2018, 5:44 PM GMT (9232e78)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -569,6 +569,18 @@ the <a>string</a> with their corresponding <a>code point</a> in <a>ASCII upper a
 <a>ASCII lowercase</a> of <var>B</var>.
 <!-- TODO: define string equals? -->
 
+<p>To <dfn export>ASCII encode</dfn> a <a>string</a> <var>input</var>, run these steps:</p>
+
+<ol>
+ <li><p><a>Assert</a>: <var>input</var> is an <a>ASCII string</a>.
+
+ <p class=note>Note: This precondition ensures that <a>isomorphic encode</a> and
+ <a>utf-8 encode</a> return the same <a>byte sequence</a> for this input.
+
+ <li><p>Return the <a>isomorphic encoding</a> of <var>input</var>.
+</ol>
+
+
 <hr>
 
 <p>To <dfn export>strip newlines</dfn> from a <a>string</a>, remove any U+000A LF and U+000D CR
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -569,6 +569,29 @@ the <a>string</a> with their corresponding <a>code point</a> in <a>ASCII upper a
 <a>ASCII lowercase</a> of <var>B</var>.
 <!-- TODO: define string equals? -->
 
+<p>To <dfn export>ASCII encode</dfn> a <a>string</a> <var>input</var>, run these steps:</p>
+
+<ol>
+ <li><p><a>Assert</a>: <var>input</var> is an <a>ASCII string</a>.
+
+ <p class=note>Note: This precondition ensures that <a>isomorphic encode</a> and
+ <a>utf-8 encode</a> return the same <a>byte sequence</a> for this input.
```
domenic: Nit: looks like the source material uppercases UTF-8. Here and below.

infra.bs
```diff
@@ -569,6 +569,29 @@ the <a>string</a> with their corresponding <a>code point</a> in <a>ASCII upper a
 <a>ASCII lowercase</a> of <var>B</var>.
 <!-- TODO: define string equals? -->
 
+<p>To <dfn export>ASCII encode</dfn> a <a>string</a> <var>input</var>, run these steps:</p>
```
domenic: Nit: remove `</p>` here and below. (I know the rules are inconsistent between HTML and other specs -_-.)

annevk:
FWIW, I think I commented on IRC that I think this is a useful concept to have as the low-level languages used to implement clients can make good use of this kind of distinction (between "ascii" and "isomorphic") for various optimizations. (And making them uncover all these possible distinctions themselves is not great.)

infra.bs
jyasskin: Whoops, thanks.

infra.bs
jyasskin: Whoops, thanks.

PR #241, opened 2019-03-25
Add flat append

### LINKED ISSUE #238: Consider defining concatenation of lists
I wanted to do that in IDL.

annevk:
To insert the items of the first given list at the end of the second given list.

Fixes #238.

***
<a href="https://whatpr.org/infra/241.html" title="Last updated on Mar 29, 2019, 8:42 AM UTC (bc36ef4)">Preview</a> | <a href="https://whatpr.org/infra/241/b49b463...bc36ef4.html" title="Last updated on Mar 29, 2019, 8:42 AM UTC (bc36ef4)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -802,10 +802,18 @@ add the given <a for=list>item</a> to the end of the list.
 <p>To <dfn export for=list>extend</dfn> a <a>list</a> |A| with a <a>list</a> |B|,
 <a for=list>for each</a> |item| of |B|, <a for=list>append</a> |item| to |A|.
 
-<p class=example id=example-list-extend>To <a for=list>extend</a> « "<code>Erin Gilbert</code>",
-"<code>Abby Yates</code>" » with « "<code>Jillian Holtzmann</code>", "<code>Patty Tolan</code>" »
-gives « "<code>Erin Gilbert</code>", "<code>Abby Yates</code>", "<code>Jillian Holtzmann</code>",
-"<code>Patty Tolan</code>" ».
+<div class=example id=example-list-extend>
+ <ol>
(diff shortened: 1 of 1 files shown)


Ms2ger:
"extend", maybe?

annevk:
That works, though it wouldn't generalize to other operations. @domenic?

domenic:
I like extend more; I think for widely-used things like this readability/alignment with convention is better than generality.

annevk:
Fine with me, added an example too.

infra.bs
```diff
@@ -799,6 +799,14 @@ out-of-bounds, except when used with <a for=list>exists</a>.
 <p>To <dfn export for=list>append</dfn> to a <a>list</a> that is not an <a>ordered set</a> is to
 add the given <a for=list>item</a> to the end of the list.
 
+<p>To <dfn export for=list>extend</dfn> a <a>list</a> |A| with a <a>list</a> |B|,
+<a for=list>for each</a> |item| of |B|, <a for=list>append</a> |item| to |A|.
+
+<p class=example id=example-list-extend>To <a for=list>extend</a> « "<code>Erin Gilbert</code>",
+"<code>Abby Yates</code>" » with « "<code>Jillian Holtzmann</code>", "<code>Patty Tolan</code>" »
+gives « "<code>Erin Gilbert</code>", "<code>Abby Yates</code>", "<code>Jillian Holtzmann</code>",
```
domenic: I think "gives" makes it a big unclear that this mutates in place. I'm OK keeping it as-is, since the normative text is right above and is pretty clear. But you could instead do an example that uses named variables and then asserts that the size is 4 after the extend operation.

annevk:
Thanks!

Issue #159, opened 2017-10-04
Further JSON processing


annevk:
Context:

* https://github.com/WICG/origin-policy/issues/32
* https://github.com/w3c/manifest/issues/611

Example of the complexity needed today:

* https://w3c.github.io/payment-method-manifest/#validate-and-parse

Folks involved: @mgiuca @tobie @marcoscaceres @kenchris @domenic @zcorpan @dhausknecht 

Not having to deal with JavaScript seems like a big step up and something we should cover. Basically convert the output to IDL/Infra types. Note that we'll have to keep the existing hook that exposes just JavaScript, as that's what Fetch and XMLHttpRequest need.

Then there's the question if we should cover more and what kind of model we want. I basically think what we want universally is to ignore a field if it's value is an unknown type. And we don't want 123 to become "123" if the field is of type string (IDL's behavior for JavaScript).

It seems like offering some kind of validation language could be useful as otherwise everyone has to reinvent it which opens up potential differences in processing, which is unlikely what web platform consumers want.

annevk:
> And we don't want 123 to become "123" if the field is of type string (IDL's behavior for JavaScript).

Note that if we do accept this, we could probably reuse most of the dictionary infrastructure here.

cc @mnot @mikewest

domenic:
I'll want this infrastructure for https://github.com/domenic/package-name-maps and will probably give a shot at generalizing it while I work on that.

I don't think we want to allow type conversions.

annevk:
https://www.w3.org/TR/tracking-dnt/#status-representation is another example of something browsers might consume.

domenic:
I think there are potentially two issues here:

1. How to write normative spec text that parses JSON into spec-level data structures (infra ordered maps/lists, plus the string/number/null/boolean primitives (see #87))
2. How to express the _expected structure_ of your JSON, and validate that the parsed JSON conforms to that structure.

(1) I think is straightforward: we basically use the existing parse JSON operation which gives a JS object, then add some stuff in Infra that crawls the JS object and produces spec-level data structures from it. Then no other spec ever has to worry about JS objects; Infra just abstracts that away into a simple bytes/string -> spec-level data structures operation.

(2) is trickier, and perhaps belongs in a separate issue. For simple cases, you can just check in prose; this will probably be necessary anyway in most cases (e.g. checking that the contents of a string are in the expected format).

[...]

PR #249, opened 2019-05-06
Add "parse JSON into maps and lists"

domenic:
Part of #159. This provides the basic parsing framework, although no validation or type-checking.

Questions:

- Is this name reasonable?
- Should we take bytes or a string as input? Offer both variants? For import maps I need a string input, so I want that available at least... But it's a bit asymmetric with the byte-accepting variant previous.

Comment for @annevk: I used Bikeshed shorthands here because I find them way more pleasant than all the close and open tags for `<a>` and `<var>`. I was tempted to use Bikeshed's built-in list and paragraph support as well, but I refrained. I realize this is inconsistent with the rest of the source text, and will change it if you so desire. I just urge you to look at the beauty that is the source of [the import maps spec](https://github.com/WICG/import-maps/blob/2f8d0fcd5becd4e6398c80530f5224fdc1d93511/spec.bs) and consider what could be :).

***
<a href="https://whatpr.org/infra/249.html" title="Last updated on May 10, 2019, 3:43 PM UTC (2739ae5)">Preview</a> | <a href="https://whatpr.org/infra/249/b522298...2739ae5.html" title="Last updated on May 10, 2019, 3:43 PM UTC (2739ae5)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -1216,8 +1216,7 @@ the involved objects or arrays are tied to a particular <a lt="realm">JavaScript
 standards, it is often more convenient to parse JSON into realm-independent <a>maps</a>,
 <a>lists</a>, <a>strings</a>, <a>booleans</a>, numbers, and nulls.
 
-<p>To <dfn export>parse JSON into maps and lists</dfn>, given a <a>string</a>
-<var>jsonText</var>:
+<p>To <dfn export>parse JSON into Infra values</dfn>, given a <a>string</a> <var>jsonText</var>:
 
 <ol>
  <li><p>Let |jsValue| be ? [$Call$](<a>%JSONParse%</a>, undefined, « |jsonText| »).
@@ -1276,6 +1275,7 @@ given a JavaScript value <var>jsValue</var>:
 
  <li><p>Return |result|.
 </ol>
+
 
 <h2 id=forgiving-base64>Forgiving base64</h2>
 
--- a/infra.bs
+++ b/infra.bs
@@ -14,8 +14,8 @@ urlPrefix: https://tc39.github.io/ecma262/#; spec: ECMA-262;
         text: %JSONStringify%; url: sec-json.stringify
         text: List; url: sec-list-and-record-specification-type
         text: The String Type; url: sec-ecmascript-language-types-string-type
-    type: method; for: Array; text: sort(); url: sec-array.prototype.sort
         text: realm; url: realm
+    type: method; for: Array; text: sort(); url: sec-array.prototype.sort
     type: abstract-op;
         text: Call; url: sec-call
         text: Get; url: sec-get-o-p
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -1074,35 +1081,99 @@ as 200/`<code>OK</code>`.
...
 </ol>
 
+<hr>
+
+<p>The above two operations operate on JavaScript values directly; in particular, this means that
+the involved objects or arrays are tied to a particular <a lt="realm">JavaScript realm</a>. In
+standards, it is often more convenient to parse JSON into realm-independent <a>maps</a>,
+<a>lists</a>, <a>strings</a>, <a>booleans</a>, numbers, and nulls.
+
+<p>To <dfn export>parse JSON into maps and lists</dfn>, given a <a>string</a>
```
annevk: It's weird that this also returns primitives. "parse JSON into Infra values" might be good enough?

infra.bs
```diff
@@ -1074,35 +1081,99 @@ as 200/`<code>OK</code>`.
...
+
+    <li><p>Let |infraValueAtKey| be the result of [=converting a JSON-derived JavaScript value to an Infra value=],
+    given |jsValueAtKey|.
+
+    <li><p>[=map/Set=] |result|[|key|] to |infraValueAtKey|.
+   </ol>
+ </li>
+
+ <li><p>Return |result|.
+</ol>
```
annevk: Let's add another newline.

marcoscaceres:
>  @marcoscaceres might also appreciate this 

Indeed... I quite like this!  

> although I think he mentioned wanting to define things in terms of JavaScript somewhere.

With this PR, I don't think doing the JS conversion is necessary, anymore - because the infra types map quite ergonomically to JS types and to C++ types. Concretely, for the use case I have in mind (web manifest), the Gecko implementation is in JS, but Chrome's implementation is in C++. Thus, Infra will  give us language/type neutrality, which is a huge plus.     

The things we then build onto of this can then operate  on the infra types 👍🔥👍

infra.bs
```diff
@@ -1074,35 +1081,99 @@ as 200/`<code>OK</code>`.
...
+
+<p>The above two operations operate on JavaScript values directly; in particular, this means that
+the involved objects or arrays are tied to a particular <a lt="realm">JavaScript realm</a>. In
+standards, it is often more convenient to parse JSON into realm-independent <a>maps</a>,
+<a>lists</a>, <a>strings</a>, <a>booleans</a>, numbers, and nulls.
+
+<p>To <dfn export>parse JSON into Infra values</dfn>, given a <a>string</a> <var>jsonText</var>:
+
+<ol>
+ <li><p>Let |jsValue| be ? [$Call$](<a>%JSONParse%</a>, undefined, « |jsonText| »).
```
annevk: I wonder if it would be more useful for callers if this returned failure upon throwing.

domenic: I think it's easier if callers can just re-throw the specific JSON parsing error, instead of having to remember that it's a SyntaxError, and using prose to stitch the error messages together. At least, that is what would work well for me in import maps.

annevk: Okay, I was wondering if maybe returning failure would be better so you don't have to catch, but if you want to re-throw that makes sense.

PR #250, opened 2019-05-17
Revise the way otherwise is supposed to be used

### LINKED ISSUE #247: Ternary revisited
#175 added a ternary construct, but we don't use it consistently. Some examples:

> Let activationTarget be target, if isActivationEvent is true and target has activation behavior, and null otherwise.

> To pop from a stack is to remove its last item and return it, if the stack is not empty, or to return nothing otherwise.

I created a Twitter poll to help https://twitter.com/annevk/status/1123219698198482944 but that didn't really get conclusive. There was some a suggestion of using neither "or" nor "and", but applying this doesn't work consistently. This maybe works:

> Let activationTarget be target, if isActivationEvent is true and target has activation behavior, null otherwise.

This does not seem to work:

> To pop from a stack is to remove its last item and return it, if the stack is not empty, to return nothing otherwise.

cc @othermaciej @gsnedders @padenot

annevk:
This also clarifies how "get the value of an entry" for an ordered map actually works.

Closes #247.

***
<a href="https://whatpr.org/infra/250.html" title="Last updated on May 20, 2019, 2:34 PM UTC (d75e378)">Preview</a> | <a href="https://whatpr.org/infra/250/9bf7638...d75e378.html" title="Last updated on May 20, 2019, 2:34 PM UTC (d75e378)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -1073,7 +1073,8 @@ interoperability among implementations.
 <p>A literal syntax can be used to express <a>ordered maps</a>, by surrounding the ordered map with
 «[ ]» characters, denoting each of its <a for=map>entries</a> as |key| → |value|, and separating its
 entries with a comma. An indexing syntax can be used to look up and set <a for=map>values</a> by
-providing a <a for=map>key</a> inside square brackets.
+providing a <a for=map>key</a> inside square brackets. The index cannot be out-of-bounds, except
+when used with <a for=map>exists</a>.
 
 <p class=example id=example-map-notation>Let |example| be the <a>ordered map</a> «[
 "<code>a</code>" → `<code>x</code>`, "<code>b</code>" → `<code>y</code>` ]». Then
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -167,8 +167,8 @@ over.
 </div>
 
 <p class=example id=example-variable-ternary>Let <var ignore>activationTarget</var> be
-<var ignore>target</var>, if <var ignore>isActivationEvent</var> is true and target has activation
-behavior, and null otherwise.
+<var ignore>target</var> if <var ignore>isActivationEvent</var> is true and target has activation
+behavior, otherwise null.
```
domenic: As @sideshowbarker noted this should be a semicolon, not a comma.

infra.bs
```diff
@@ -1082,9 +1082,9 @@ providing a <a for=map>key</a> inside square brackets.
 <hr>
 
 <p>To <dfn export for=map lt="get|get the value">get the value of an entry</dfn> in an
-<a>ordered map</a> given a <a for=map>key</a> is to retrieve the <a for=map>value</a> of any
-existing <a for=map>entry</a> if the map <a for=map>contains</a> an entry with the given key, or
-to return nothing otherwise. We can also use the indexing syntax explained above.
+<a>ordered map</a> given a <a for=map>key</a>: if the map <a for=map>contains</a> an entry with the
+given key, then return the <a for=map>value</a> of the matching <a for=map>entry</a>; otherwise,
+return nothing. We can also use the indexing syntax explained above.
```
domenic: If we're changing this, I think we should actually just assert that it contains it. This "nothing" value is confusing. For lists we say

> The index cannot be out-of-bounds, except when used with exists.

infra.bs
annevk: That was in reaction to the ", null otherwise" variant. Also, it's not the style we used in https://infra.spec.whatwg.org/#code-unit-prefix.

infra.bs
annevk: Yeah, that sounds good.

PR #252, opened 2019-05-23
Editorial: improve code unit prefix

annevk:
***
<a href="https://whatpr.org/infra/252.html" title="Last updated on May 23, 2019, 9:04 AM UTC (211e5ab)">Preview</a> | <a href="https://whatpr.org/infra/252/e65a4e0...211e5ab.html" title="Last updated on May 23, 2019, 9:04 AM UTC (211e5ab)">Diff</a>


annevk:
Well, for the existing code the value of _bCodeUnit_ does not matter for these steps. For both null and non-null, if _aCodeUnit_ is null, we return true.

For your example, _b_ is a code unit prefix of _a_ will return true, so it'll be false. (It's not affected by these changes as far as I can tell.)

Issue #206, opened 2018-06-22
Define "steps"


jakearchibald:
Some algorithms take a set of steps as a find of "callback", such as "Queue a task". Should this argument type be defined by infra?

annevk:
Yeah, maybe. It's effectively an algorithm so we should maybe make those synonyms or some such?

skitzofrenicdev:
Pull

PR #255, opened 2019-06-14
Add and explain usage of "abort these steps"

domenic:
Related to #206. Stems from https://github.com/WICG/portals/pull/138#discussion_r292641909.

/cc @jeremyroman

***
<a href="https://whatpr.org/infra/255.html" title="Last updated on Jun 14, 2019, 6:40 PM UTC (b92aede)">Preview</a> | <a href="https://whatpr.org/infra/255/ac90e66...b92aede.html" title="Last updated on Jun 14, 2019, 6:40 PM UTC (b92aede)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -183,6 +183,33 @@ will make the caller automatically rethrow the given value, if any, and thereby 
 caller's algorithm. Using prose the caller has the ability to "catch" the exception and perform
 another action.
 
+<p>Algorithms often involve passing a set of steps as an argument to some other algorithm. In
+these cases, use "abort these steps" to indicate that the passed set of steps should terminate,
+without affecting the larger algorithm. Do not use "return" for this, as "return" is only used for
+terminating the larger algorithm.
+
+<div class=example id=example-abort-these-steps>
+  <p>To bake a cake:
+
+  <ol>
+   <li><p>Gather the ingredients.
+   <li><p>If you are missing crucial ingredients, then return.
+   <li><p>Preheat the oven.
+   <li>
+    <p>Ask your partner to complete the following steps:
+    <ol>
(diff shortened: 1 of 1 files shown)


annevk:
The example is a bit unclear to me as none of the terms are defined. Is "ask your partner" an algorithm? Could you use this inline, e.g., "Run these steps:" and then "abort these steps" would not terminate the outer set of steps? I'm not sure that matches all existing usage.

GPHemsley:
Indeed, "ask your partner" implies passing information in one direction, but there is no subsequent step passing information back in the other direction. How do you know when your partner is done? Or whether they were even successful in completing the steps? (That is, what happens if they abort the steps, as requested?)

And then later on, the main list of steps mentions batter, but that is undefined. Was the partner making the batter? Should we check to make sure the batter is ready before we pour it into the pan?

And as this is the algorithm to "bake a cake", if at any point we "return" or "abort these steps", we have not baked a cake, which seems like a Bad Thing™.

More broadly, an algorithm is a set of steps. Just because you sometimes pass a set of steps into another set of steps doesn't make either set of steps any less of an algorithm. So you might want to differentiate between an intentionally-incomplete set of steps (i.e., some sort of mix-in) from a nested-but-complete set of steps (like a direct function call or a callback argument), and then the language of how to exit the set of steps (either prematurely or at the end) might fall out from that (I hope).

domenic:
I'm discouraged by how this example isn't as clear as I thought, and will not really have time to pursue this further. (It was supposed to be "I have a couple hours free and I'd like to work on something simple"). As such I'll close this PR and let others who have stronger opinions about baking cakes and partners and such work on this issue.

PR #259, opened 2019-06-26
Fix "starts with" and "code unit prefix" algorithm

### LINKED ISSUE #256: Incorrect algorithm for byte sequence starts with and string code unit prefix
Byte sequence **starts with** https://infra.spec.whatwg.org/#byte-sequence-starts-with has a step (2.3 "If aByte is null, then return true") that results in treating certain shorter sequences as starting with certain longer sequences, e.g. `abc` starting with `abcd`, while really it's vice versa.

String **code unit prefix** https://infra.spec.whatwg.org/#code-unit-prefix, being essentialy the same algorithm, has the same problem.

al-arz:
Closes #256.

Some test cases to consider:

```
"abcd".startsWith("") // true
"abcd".starts("a") // true
"abcd".startsWith("abcd") // true
"abcd".startsWith("abcde") // false
"abcd".startsWith("abd") // false
"".startsWith("") // true
"".startsWith("abcd") // false
```

***
<a href="https://whatpr.org/infra/259.html" title="Last updated on Jun 27, 2019, 7:01 PM UTC (bfae6fa)">Preview</a> | <a href="https://whatpr.org/infra/259/ac90e66...bfae6fa.html" title="Last updated on Jun 27, 2019, 7:01 PM UTC (bfae6fa)">Diff</a>


al-arz:
Well, why not? :)

PR #269, opened 2019-09-12
Define "user agent".

jyasskin:
This definition draws from previous definitions in RFC 2616,
https://www.w3.org/TR/UAAG20/#def-user-agent,
https://www.w3.org/TR/CSS2/conform.html#user-agent, and
https://en.wikipedia.org/wiki/User_agent.

I am not at all tied to the particular wording of this definition or the choice of specification it appears in. If it moves to HTML, it can probably be more specific, maybe along the lines of https://www.w3.org/TR/CSS2/conform.html#user-agent.

Fixes whatwg/html#1021.

***
<a href="https://whatpr.org/infra/269.html" title="Last updated on Sep 13, 2019, 5:33 PM UTC (cafeaff)">Preview</a> | <a href="https://whatpr.org/infra/269/d109ae2...cafeaff.html" title="Last updated on Sep 13, 2019, 5:33 PM UTC (cafeaff)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -98,7 +98,7 @@ either width or height is zero"), means an inclusive "or" (implying "or both"), 
 out as being exclusive (with "but not both").
 
 <p>A <dfn export>user agent</dfn> is any software that acts on behalf of a user, for example by
-retrieving and rendering Web content and facilitating end user interaction with it. In
+retrieving and rendering web content and facilitating end user interaction with it. In
 specifications using the Infra Standard, the user agent is generally the client software that
 implements the specification.
 
--- a/infra.bs
+++ b/infra.bs
@@ -97,6 +97,11 @@ violation.
 either width or height is zero"), means an inclusive "or" (implying "or both"), unless it is called
 out as being exclusive (with "but not both").
 
+<p>A <dfn export>user agent</dfn> is any software that acts on behalf of a user, for example by
+retrieving and rendering Web content and facilitating end user interaction with it. In
+specifications using the Infra Standard, the user agent is generally the client software that
+implements the specification.
+
 
 <h2 id=algorithms>Algorithms</h2>
 


GPHemsley:
> Generally looks good. I'm a bit unsure about the "for example" connector, but probably it's the best way to go.

s/for example/such as/ ?

infra.bs
```diff
@@ -97,6 +97,11 @@ violation.
 either width or height is zero"), means an inclusive "or" (implying "or both"), unless it is called
 out as being exclusive (with "but not both").
 
+<p>A <dfn export>user agent</dfn> is any software that acts on behalf of a user, for example by
+retrieving and rendering Web content and facilitating end user interaction with it. In
```
annevk: Lowercase web per https://whatwg.org/style-guide.

jyasskin: Whoops, thanks, done.

annevk:
Merge this after TPAC when there's no more feedback?

infra.bs
```diff
@@ -97,6 +97,11 @@ violation.
 either width or height is zero"), means an inclusive "or" (implying "or both"), unless it is called
 out as being exclusive (with "but not both").
 
+<p>A <dfn export>user agent</dfn> is any software that acts on behalf of a user, for example by
```
sideshowbarker:
```suggestion
<p>A <dfn export>user agent</dfn> is any software that acts on behalf of a user; for example by
```

jyasskin: The interwebs say to use a semicolon here (and a comma after) when the second bit is a complete sentence, which it isn't here. I don't feel strongly, of course.

sideshowbarker: > The interwebs say to use a semicolon here (and a comma after) when the second bit is a complete sentence, which it isn't here. I don't feel strongly, of course.

I don’t feel strongly about it either — so it’s fine by me to leave it as-is

GPHemsley: I still think s/for example/such as/ is a better approach.

annevk:
I kept it as "for example by" since "such as by" seemed awkward to me. Different approaches welcome though as follow-up PRs.

PR #289, opened 2020-02-03
Change code unit-dependent algorithms to operate on JavaScript strings.

### LINKED ISSUE #288: Code unit prefix uses the wrong string length
For the longest time I was under the impression that the definitions of length for Javascript strings and for (not explicitly Javascript) strings were equivalent. The difference is easy to miss when skimming through the spec, but what makes it clear that the definition can be confusing is that the code unit prefix algorithm just below uses the wrong kind of length:

> Let `aCodeUnit` be the `i`th [code unit](https://infra.spec.whatwg.org/#code-unit) of `a` if `i` is less than `a`’s [length](https://infra.spec.whatwg.org/#string-length); otherwise null.

"Length" there links to the definition of a (not explicitly Javascript) string's length, which measures code points, not code units. Aside from changing the link, the text would also need to be updated, though I'm not sure if something like "`a`'s length as a Javascript string" would be the way to do it.

I'd also ask for a note clarifying the difference between the lengths, but that can be its own separate issue if need be.

andreubotella:
This change updates the code unit prefix and code unit less than algorithms to operate on JavaScript strings, rather than strings, since they only operate on them as sequences of code units. Additionally, the code unit prefix algorithm mistakenly depends on the string's length in a context when it needs to measure the number of code units, which this change replaces to the JavaScript string's length.

Fixes #288.

***
<a href="https://whatpr.org/infra/289.html" title="Last updated on Feb 3, 2020, 11:26 AM UTC (0d9420a)">Preview</a> | <a href="https://whatpr.org/infra/289/d3549e1...0d9420a.html" title="Last updated on Feb 3, 2020, 11:26 AM UTC (0d9420a)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -663,9 +663,9 @@ implementations of just <a>JavaScript strings</a> for performance and memory rea
 
 <hr>
 
-<p>A <a>string</a> <var>a</var> is a
-<dfn export lt="code unit prefix|starts with">code unit prefix</dfn> of a <a>string</a> <var>b</var>
-if the following steps return true:
+<p>A <a>JavaScript string</a> <var>a</var> is a
+<dfn export lt="code unit prefix|starts with">code unit prefix</dfn> of a <a>JavaScript string</a>
+<var>b</var> if the following steps return true:
 
 <ol>
  <li><p>Let <var>i</var> be 0.
@@ -675,10 +675,10 @@ if the following steps return true:
 
   <ol>
    <li><p>Let <var>aCodeUnit</var> be the <var>i</var>th <a>code unit</a> of <var>a</var> if
-   <var>i</var> is less than <var>a</var>'s <a for=string>length</a>; otherwise null.
+   <var>i</var> is less than <var>a</var>'s <a for="JavaScript string">length</a>; otherwise null.
 
    <li><p>Let <var>bCodeUnit</var> be the <var>i</var>th <a>code unit</a> of <var>b</var> if
-   <var>i</var> is less than <var>b</var>'s <a for=string>length</a>; otherwise null.
+   <var>i</var> is less than <var>b</var>'s <a for="JavaScript string">length</a>; otherwise null.
 
    <li><p>If <var>bCodeUnit</var> is null, then return true.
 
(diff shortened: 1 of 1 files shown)


annevk:
I'd like some more discussion then.

Is the code unit length of a scalar value string derived through automatic conversion into a JavaScript string?

Because we automatically convert in this direction, there's no real need to distinguish between both string types in the header of an algorithm, especially since code units are a bit of an implementation detail for scalar value strings in this case?

(Only code point strings could create trouble, but we don't have those.)

andreubotella:
Since only "Javascript strings" are defined as a sequence of code units, it seems like any usage of code units must be in the context of a "Javascript string", even if through an implicit conversion. In that sense, if you need the string's code unit length, either you declare it as a Javascript string (as at the top of the algorithm here) or you explicitly convert to it in the algorithm (for example, "`a`'s length as a Javascript string", as I suggested in #288). Simply using "length" in the context of a "string" would be inappropriate, even if linked to the "Javascript string" definition.

But it seems like it would be much more intuitive to rename the lengths to make what they're measuring explicit: "code unit length" versus "code point length". And maybe keep the term "length" for BMP-only cases. But that would require adjusting all callers.

domenic:
Taking a step back, I think we need to ask what direction we want to guide the specification ecosystem in. I see two possible end points:

1. The specification ecosystem generally uses "string", which is basically an alias for "JavaScript string". Sometimes it uses "scalar value string", a particular type of (JavaScript) string, when it wants to make it clear that no unpaired surrogates exist. (E.g. "Assert: _s_ is a scalar value string"). All operations apply to strings, in general.

2. The specification ecosystem carefully ensures they are using a "JavaScript string" or a "scalar value string", and uses the appropriate operation for each of them. This leads to, e.g., defining a separate prefix or less-than operation for the two types, and considering the bare term "string" deprecated.

I thought we were going in the direction of (1) with Infra. This PR seems like a slight nudge in the direction of (2), which is why I find it questionable. On the other hand, the fact that we have separate "length" definitions means maybe we are already trending in the direction of (2).

andreubotella:
I see. I originally suggested in #288 to use the text "`a`'s length as a Javascript string", which although it can be interpreted as a type conversion, it can simply be seen as pointing to what definition of "length" to use. @annevk preferred simply changing the link without updating the text (which I don't agree with) and suggested making the type a Javascript string in the header of "code unit prefix" to force that interpretation. I then extended that to "code unit less than", which doesn't use length but does use code units, for consistency.

I would be fine with staying in the direction of (1) by updating the text to force an interpretation of code unit length without making those algorithms explicitly Javascript string algorithms. Some other pull request could then deal with possibly renaming the two length definitions.

annevk:
This would be a lot easier if we could find all dependencies.

I think I prefer equating string and JavaScript string, defining scalar value string and ASCII string as subsets, and then have code unit length and code point length. One of code unit length or code point length should also be called length, but which would really depend on usage.

andreubotella:
Since the changes requested by the reviewers go in a different direction than this PR was meant to go, I'm closing it and opening a new one for the same issue that simply changes the use of "length" in the code unit prefix algorithm.

PR #290, opened 2020-02-11
Fix the usage of "length" in the code unit prefix algorithm

### LINKED ISSUE #288: Code unit prefix uses the wrong string length

andreubotella:
This change corrects the usage of "length" as applied to strings in the code unit prefix algorithm, when in context it measures the number of code units, not code points.

Fixes #288.

***
<a href="https://whatpr.org/infra/290.html" title="Last updated on Feb 12, 2020, 1:24 AM UTC (41f3d7e)">Preview</a> | <a href="https://whatpr.org/infra/290/d3549e1...41f3d7e.html" title="Last updated on Feb 12, 2020, 1:24 AM UTC (41f3d7e)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -675,10 +675,10 @@ if the following steps return true:
 
   <ol>
    <li><p>Let <var>aCodeUnit</var> be the <var>i</var>th <a>code unit</a> of <var>a</var> if
-   <var>i</var> is less than <var>a</var>'s <a for=string>length</a>; otherwise null.
+   <var>i</var> is less than <var>a</var>'s <a for="JavaScript string">length</a>; otherwise null.
 
    <li><p>Let <var>bCodeUnit</var> be the <var>i</var>th <a>code unit</a> of <var>b</var> if
-   <var>i</var> is less than <var>b</var>'s <a for=string>length</a>; otherwise null.
+   <var>i</var> is less than <var>b</var>'s <a for="JavaScript string">length</a>; otherwise null.
 
    <li><p>If <var>bCodeUnit</var> is null, then return true.
 
@@ -1495,6 +1495,7 @@ certain inputs.
 
 <p>Many thanks to
 Addison Phillips,
+Andreu Botella,
 Aryeh Gregor,
 Chris Rebert,
 Daniel Ehrenberg,


andreubotella:
This is a replacement for pull request #289.

annevk:
I don't think we had agreement on "as a JavaScript string" unfortunately. Maybe a good initial take here is to fix the reference and then file a follow-up for:

> I think I prefer equating string and JavaScript string, defining scalar value string and ASCII string as subsets, and then have code unit length and code point length. One of code unit length or code point length should also be called length, but which would really depend on usage.

The follow-up should be easier once there's a global tool to look at usage of terms. Though I suppose we could start by having specific terms and using those as applicable.

infra.bs
```diff
@@ -675,10 +675,12 @@ if the following steps return true:
 
   <ol>
    <li><p>Let <var>aCodeUnit</var> be the <var>i</var>th <a>code unit</a> of <var>a</var> if
-   <var>i</var> is less than <var>a</var>'s <a for=string>length</a>; otherwise null.
+   <var>i</var> is less than <var>a</var>'s <a for="JavaScript string">length</a> as a
+   <a>JavaScript string</a>; otherwise null.
```
domenic: Please remove "as a JavaScript string"; as discussed this just confuses matters.

andreubotella:
I'm not sure about having a term which is ambiguous in context distinguished only by which definition it's pointing to. But I do agree with @annevk about filing a follow-up issue, so I guess it'll do as a temporary fix.

annevk:
Thanks for keeping this pragmatic! Filed #291 on string improvements.

Issue #291, opened 2020-02-12
Clean up strings and string lengths


annevk:
As discussed in #289 the current setup can be confusing. This is what I think we should do:

> I think I prefer equating string and JavaScript string, defining scalar value string and ASCII string as subsets, and then have code unit length and code point length. One of code unit length or code point length should also be called length, but which would really depend on usage.

A reasonable start might be to do the above and keep both lengths we have today around as aliases for code unit length and code point length until we know which one to deprecate and remove. (Or perhaps we should favor neither and try to remove both shorthands.)

cc @andreubotella

andreubotella:
I've been looking through the standards listed in the wiki as dependent on Infra, to inform how this issue proceeds and where changes are needed. Though we should also do some reasonable effort to check the usages on dependent specs that aren't listed in the wiki.

In general, specifications use the term "string" a lot, usually not bothering to reference Infra (and when they do it's once or twice in the whole spec). But "JavaScript string" is more rare. Below I'm listing all cases where "JavaScript string" is used, as well as any usages of either "length". Dependent specs I'm not listing aren't affected:

HTML:
- Beside the many times that HTML uses "string" without linking back to Infra, it also uses "JavaScript string" ([1](https://html.spec.whatwg.org/multipage/dom.html#the-innertext-idl-attribute:javascript-string), [2](https://html.spec.whatwg.org/multipage/webappapis.html#creating-scripts:javascript-string), [3](https://html.spec.whatwg.org/multipage/webappapis.html#creating-scripts:javascript-string-2), [4](https://html.spec.whatwg.org/multipage/webappapis.html#integration-with-the-javascript-module-system:javascript-string), [5](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-for-spec-authors:javascript-string), [6](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-for-spec-authors:javascript-string-2), [7](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-for-spec-authors:javascript-string-3)). There doesn't seem to be a pattern to the usage of one term over the other.
[...]

PR #292, opened 2020-02-12
Redo JavaScript and scalar value strings

annevk:
This changes things a bit so there's one clear base type, called string. A scalar value string is now defined as a subset, similar to ASCII string.

A string has two different lengths, a code unit length and a code point length, and for now we keep the legacy length dfn (for JavaScript string and string) as we're not sure of all dependencies.

Fixes part of #291.

cc @andreubotella

***
<a href="https://whatpr.org/infra/292.html" title="Last updated on Mar 10, 2020, 8:57 AM UTC (b560129)">Preview</a> | <a href="https://whatpr.org/infra/292/646a05d...b560129.html" title="Last updated on Mar 10, 2020, 8:57 AM UTC (b560129)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -628,13 +628,13 @@ them effectively as-is.
 <a>code units</a> 0xD83D, 0xDCA9, and 0xD800, when interpreted as containing <a>code points</a>,
 would consist of the <a>code points</a> U+1F4A9 and U+D800.
 
-<p>A <a>string</a>'s <dfn export for=string>code unit length</dfn> is the number of
-<a>code units</a> it contains. For now, this can also be referred to as a <a>JavaScript string</a>'s
-<dfn export for="JavaScript string">length</dfn>.
+<p>A <a>string</a>'s
+<dfn export for="string,JavaScript string,scalar value string">length</dfn> is the number of
+<a>code units</a> it contains.
 
-<p>A <a>string</a>'s <dfn export for=string>code point length</dfn> is the number of
-<a>code points</a> it contains. For now, this can also be referred to as a <a>string</a>'s
-<dfn export for=string>length</dfn>.
+<p>A <a>string</a>'s
+<dfn export for="string,JavaScript string,scalar value string">code point length</dfn> is the number
+of <a>code points</a> it contains.
 
 <p>A <dfn export>scalar value string</dfn> is a <a>string</a> whose <a>code points</a> are all
 <a>scalar values</a>.
@@ -676,10 +676,10 @@ if the following steps return true:
 
   <ol>
    <li><p>Let <var>aCodeUnit</var> be the <var>i</var>th <a>code unit</a> of <var>a</var> if
-   <var>i</var> is less than <var>a</var>'s <a for=string>code unit length</a>; otherwise null.
+   <var>i</var> is less than <var>a</var>'s <a for=string>length</a>; otherwise null.
 
    <li><p>Let <var>bCodeUnit</var> be the <var>i</var>th <a>code unit</a> of <var>b</var> if
-   <var>i</var> is less than <var>b</var>'s <a for=string>code unit length</a>; otherwise null.
+   <var>i</var> is less than <var>b</var>'s <a for=string>length</a>; otherwise null.
 
    <li><p>If <var>bCodeUnit</var> is null, then return true.
 
--- a/infra.bs
+++ b/infra.bs
@@ -629,8 +629,8 @@ them effectively as-is.
 would consist of the <a>code points</a> U+1F4A9 and U+D800.
 
 <p>A <a>string</a>'s
-<dfn export for="string,JavaScript string,scalar value string">length</dfn> is the number of
-<a>code units</a> it contains.
+<dfn export for="string,JavaScript string,scalar value string" id="javascript-string-length">length</dfn>
+is the number of <a>code units</a> it contains.
 
 <p>A <a>string</a>'s
 <dfn export for="string,JavaScript string,scalar value string">code point length</dfn> is the number
--- a/infra.bs
+++ b/infra.bs
@@ -7,10 +7,6 @@ Abstract: The Infra Standard aims to define the fundamental concepts upon which 
 Translation: ja https://triple-underscore.github.io/infra-ja.html
 </pre>
 
-<pre class=link-defaults>
-spec:infra; type:dfn; text:string
-</pre>
-
 <pre class=anchors>
 urlPrefix: https://tc39.github.io/ecma262/#; spec: ECMA-262;
     type: dfn
--- a/infra.bs
+++ b/infra.bs
@@ -7,6 +7,10 @@ Abstract: The Infra Standard aims to define the fundamental concepts upon which 
 Translation: ja https://triple-underscore.github.io/infra-ja.html
 </pre>
 
+<pre class=link-defaults>
+spec:infra; type:dfn; text:string
+</pre>
+
 <pre class=anchors>
 urlPrefix: https://tc39.github.io/ecma262/#; spec: ECMA-262;
     type: dfn
--- a/infra.bs
+++ b/infra.bs
@@ -520,7 +520,7 @@ contains, in the range 0x61 (a) to 0x7A (z), inclusive, by 0x20.
 <hr>
 
 <p>To <dfn export>isomorphic decode</dfn> a <a>byte sequence</a> <var>input</var>, return a
-<a>string</a> whose <a for=string>length</a> is equal to <var>input</var>'s
+<a>string</a> whose <a for=string>code point length</a> is equal to <var>input</var>'s
 <a for="byte sequence">length</a> and whose <a>code points</a> have the same values as
 <var>input</var>'s <a>bytes</a>, in the same order.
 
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -628,38 +628,40 @@ them effectively as-is.
...
-<p>A <dfn export>scalar value string</dfn> is a sequence of <a>scalar values</a>.
+<p>A <a>string</a>'s <dfn export for=string>code unit length</dfn> is the number of
+<a>code units</a> it contains. For now, this can also be referred to as a <a>JavaScript string</a>'s
+<dfn export for="JavaScript string">length</dfn>.
+
+<p>A <a>string</a>'s <dfn export for=string>code point length</dfn> is the number of
+<a>code points</a> it contains. For now, this can also be referred to as a <a>string</a>'s
+<dfn export for=string>length</dfn>.
+
+<p>A <dfn export>scalar value string</dfn> is <a>string</a> whose <a>code points</a> are all
```
andreubotella:
```suggestion
<p>A <dfn export>scalar value string</dfn> is a <a>string</a> whose <a>code points</a> are all
```

infra.bs
```diff
@@ -628,38 +628,40 @@ them effectively as-is.
...
+<div class=note>
+ <p>The replaced surrogates are always isolated surrogates, since the process of interpreting the
+ JavaScript string as containing <a>code points</a> will have converted surrogate pairs into
+ <a>scalar values</a>.
 
-<p>A <a>scalar value string</a> can always be used as <a>JavaScript string</a> implicitly since it
-is a subset. The reverse is only possible if the <a>JavaScript string</a> is known to not contain
-<a>surrogates</a>; otherwise a <a for="JavaScript string" lt=convert>conversion</a> must be
-performed.
+ <p>A <a>scalar value string</a> can always be used as <a>string</a> implicitly since it is a
```
andreubotella:
```suggestion
 <p>A <a>scalar value string</a> can always be used as a <a>string</a> implicitly since it is a
```

infra.bs
```diff
@@ -628,38 +628,40 @@ them effectively as-is.
...
 <a>code points</a>, would consist of the <a>code points</a> U+1F4A9 and U+D800.
 
-<p>A <dfn export>scalar value string</dfn> is a sequence of <a>scalar values</a>.
+<p>A <a>string</a>'s <dfn export for=string>code unit length</dfn> is the number of
+<a>code units</a> it contains. For now, this can also be referred to as a <a>JavaScript string</a>'s
+<dfn export for="JavaScript string">length</dfn>.
+
+<p>A <a>string</a>'s <dfn export for=string>code point length</dfn> is the number of
+<a>code points</a> it contains. For now, this can also be referred to as a <a>string</a>'s
+<dfn export for=string>length</dfn>.
```
domenic: Oof, now that this is explicit, it is really confusing. Can we pick only one that is the string's "length", and delete the other? And hope that doesn't break things in the ecosystem too badly?

I think "code unit length" is most likely to be correct since it matches JavaScript thus IDL...

infra.bs
```diff
@@ -628,38 +628,40 @@ them effectively as-is.
...
+
```
domenic: Perhaps we should add an XXX box with an issue reference to explain the "for now"?

infra.bs
```diff
@@ -628,38 +628,40 @@ them effectively as-is.
...
+ <p>The replaced surrogates are always isolated surrogates, since the process of interpreting the
+ JavaScript string as containing <a>code points</a> will have converted surrogate pairs into
+ <a>scalar values</a>.
 
-<p>A <a>scalar value string</a> can always be used as <a>JavaScript string</a> implicitly since it
-is a subset. The reverse is only possible if the <a>JavaScript string</a> is known to not contain
-<a>surrogates</a>; otherwise a <a for="JavaScript string" lt=convert>conversion</a> must be
-performed.
+ <p>A <a>scalar value string</a> can always be used as <a>string</a> implicitly since it is a
+ subset. The reverse is only possible if the <a>JavaScript string</a> is known to not contain
```
domenic: JavaScript string -> string here; the two sentences should agree

infra.bs
annevk: Yeah, discussing  this with @andreubotella I think we want "length" and "code point length" and probably use the former if it doesn't matter or is better?

The main problem is redefining "length" for string as currently that means code point length. Maybe that's okay though and this hasn't seen much adoption yet. @andreubotella was looking into that as well.

annevk:
Okay, I decided we should go for length and code point length and at least update/prepare PRs for the specifications identified in https://github.com/whatwg/infra/issues/291#issuecomment-586063726 (thanks @andreubotella). I'll work on that next.

This PR now reflects that thinking and can be reviewed again.

infra.bs
```diff
@@ -608,58 +608,58 @@ U+007A (z), inclusive.
 
 <h3 id=strings>Strings</h3>
 
-<p>A <dfn export>JavaScript string</dfn> is a sequence of unsigned 16-bit integers, also known as
-<dfn export lt="code unit">code units</dfn>.
+<p>A <dfn export>string</dfn> is a sequence of unsigned 16-bit integers, also known as
+<dfn export lt="code unit">code units</dfn>. A <a>string</a> is also known as a
+<dfn export>JavaScript string</dfn>. <a>Strings</a> are denoted by double quotes and monospace font.
```
domenic: I'm happy with this, but want to mention a couple alternatives/additions:

- State situations in which one might prefer to use "JavaScript string" instead of just "string", e.g. "A string is also known as a JavaScript string; typically this is used to emphasize the nature of the string as a collection of code units, as opposed to the below-discussed interpretation as containing code points."

- Use `<dfn export lt="string,JavaScript string">string</dfn>` and remove the `<dfn>` for JavaScript string, either converting it to a link or just leaving it without markup.

infra.bs
```diff
@@ -608,58 +608,58 @@ U+007A (z), inclusive.
...
+<div class=note>
+ <p>The replaced surrogates are always isolated surrogates, since the process of interpreting the
+ string as containing <a>code points</a> will have converted surrogate pairs into
+ <a>scalar values</a>.
 
-<p class=note>The replaced surrogates are always isolated surrogates, since the process of
-interpreting the JavaScript string as containing <a>code points</a> will have converted surrogate
-pairs into <a>scalar values</a>.
+ <p>A <a>scalar value string</a> can always be used as a <a>string</a> implicitly since it is a
+ subset. The reverse is only possible if the <a>string</a> is known to not contain
```
domenic: Preexisting issue, but "it is a subset" wording doesn't seem quite right. "A scalar value string" is not a subset of "a string", which is what this grammatical structure implies. Potential rephrasings:

- "since every scalar value string is a string"
- "since the set of all scalar value strings is a subset of the set of all strings"

infra.bs
annevk: I didn't go with "typically" since I'm not really sure what will be the dominant pattern long term.

annevk:
I plan on landing this somewhere tomorrow. If anyone wants more review time now would be a good time to say so.

infra.bs
```diff
@@ -608,58 +612,60 @@ U+007A (z), inclusive.
...
+ <p>The replaced surrogates are always isolated surrogates, since the process of interpreting the
+ string as containing <a>code points</a> will have converted surrogate pairs into
+ <a>scalar values</a>.
 
-<p>A <a>scalar value string</a> can always be used as <a>JavaScript string</a> implicitly since it
-is a subset. The reverse is only possible if the <a>JavaScript string</a> is known to not contain
-<a>surrogates</a>; otherwise a <a for="JavaScript string" lt=convert>conversion</a> must be
-performed.
+ <p>A <a>scalar value string</a> can always be used as a <a>string</a> implicitly since every
+ <a>scalar value string</a> is a <a>string</a>. The reverse is only possible if the <a>string</a> is
```
andreubotella:
```suggestion
 <a>scalar value string</a> is a <a>string</a>. The reverse is only so when the <a>string</a> is
```
"The reverse is only possible" already sounded odd before, but with this change it's particularly weird.

infra.bs
```diff
@@ -7,7 +7,11 @@ Abstract: The Infra Standard aims to define the fundamental concepts upon which
 Translation: ja https://triple-underscore.github.io/infra-ja.html
 </pre>
 
-<pre class="anchors">
+<pre class=link-defaults>
+spec:infra; type:dfn; text:string
+</pre>
```
annevk: This can be removed again tomorrow I suspect because of https://github.com/w3c/csswg-drafts/commit/ed9468f9f168a5ed6c7d2ef2cea6134bf4a63cd5.

infra.bs
annevk: I pushed a slightly more explicit variant of this.

andreubotella: LGTM

infra.bs
```diff
@@ -608,58 +612,60 @@ U+007A (z), inclusive.
...
-<p class=example id=example-string-notation>"<code>Hello, world!</code>" is a string.
-
-<p>A <a>string</a>'s <dfn export for=string>length</dfn> is the number of <a>code points</a> it
-contains.
-
-<p>To <dfn export for="JavaScript string">convert</dfn> a <a>JavaScript string</a> into a
-<a>scalar value string</a>, replace any <a>surrogates</a> with U+FFFD.
-<!-- Obviates need for https://heycam.github.io/webidl/#dfn-obtain-unicode -->
+<p>To <dfn export for="string,JavaScript string" id=javascript-string-convert>convert</dfn> a
+<a>string</a> into a <a>scalar value string</a>, replace any <a>surrogates</a> with U+FFFD.
```
aphillips: Should this say "unpaired surrogates" or "surrogate code points", since a string/JavaScript string can contain surrogate code *units*? I see that the following note covers this, but maybe calling out code points here would help?

annevk: I'm going to leave this as-is since surrogate is defined as a code point and U+FFFD is also a code point (the equivalent code unit would be denoted as 0xFFFD). If I keep getting questions about this though I'll reconsider.

PR #295, opened 2020-03-10
Meta: add another ID for string's length

annevk:
An alternative here would be to use `oldids`, but I'm not sure if that adds enough benefit for the cruft it creates.

***
<a href="https://whatpr.org/infra/295.html" title="Last updated on Mar 10, 2020, 1:02 PM UTC (f8dd58d)">Preview</a> | <a href="https://whatpr.org/infra/295/587f3c2...f8dd58d.html" title="Last updated on Mar 10, 2020, 1:02 PM UTC (f8dd58d)">Diff</a>


domenic:
I minorly prefer oldids for the semantic/understandability benefit for us when working on the document.

PR #296, opened 2020-04-01
Meta: update repository files

annevk:
See https://github.com/whatwg/spec-factory for details.

I think we probably don't want PULL_REQUEST_TEMPLATE.md here, but I figured I'd put it up to see what people think.


domenic:
+1 for removing the pull request template.

PR #303, opened 2020-05-11
Add definitions for string "is" and "identical to"

aphillips:
Fixes whatwg:html#5067

Adds definitions for string equality comparisons "is" and "identical to". Includes a blanket statement making these the default for string equality. Also includes a note spelling out the relationship to HTML's former "case-sensitive" comparison and appropriate health warning about visual and encoding sequence identity.

***
<a href="https://whatpr.org/infra/303.html" title="Last updated on May 12, 2020, 3:37 PM UTC (0b5a717)">Preview</a> | <a href="https://whatpr.org/infra/303/6ed08d3...0b5a717.html" title="Last updated on May 12, 2020, 3:37 PM UTC (0b5a717)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -679,21 +679,18 @@ where <a>UTF-8 encode</a> comes into play.
 
 <hr>
 
-<p>A <a>string</a> <var>a</var> <dfn export for=string lt="is|identical to">is</dfn> or is 
-<a for=string>identical to</a> a <a>string</a> <var>b</var> if it consists of the same sequence of 
+<p>A <a>string</a> <var>a</var> <dfn export for=string lt="is|identical to">is</dfn> or is
+<a for=string>identical to</a> a <a>string</a> <var>b</var> if it consists of the same sequence of
 <a>code units</a>.
 
 <p>Except where otherwise stated, all string comparisons use <a for=string>is</a>.
 
-<div class=note>
-
- <p>This type of <a>string</a> comparison was formerly known as a "case-sensitive"
- comparison in HTML. Strings that compare as <a>identical to</a> one another are not only
- sensitive to case variation (such as UPPER and lower case), but also to other code point encoding 
- choices, such as normalization form or the order of combining marks. Two strings that are visually or
- even canonically equivalent according to Unicode might still not be <a>identical to</a>
- each other.
-</div>
+<p class=note>This type of <a>string</a> comparison was formerly known as a "case-sensitive"
+comparison in <cite>HTML</cite>. Strings that compare as <a>identical to</a> one another are not
+only sensitive to case variation (such as UPPER and lower case), but also to other code point
+encoding  choices, such as normalization form or the order of combining marks. Two strings that are
+visually or even canonically equivalent according to Unicode might still not be
+<a for=string>identical to</a> each other. [[HTML]] [[UNICODE]]</p>
 
 <p>A <a>string</a> <var>a</var> is a
 <dfn export lt="code unit prefix|starts with">code unit prefix</dfn> of a <a>string</a> <var>b</var>
--- a/infra.bs
+++ b/infra.bs
@@ -686,10 +686,10 @@ where <a>UTF-8 encode</a> comes into play.
 <p>Except where otherwise stated, all string comparisons use <a for=string>is</a>.
 
 <p class=note>This type of <a>string</a> comparison was formerly known as a "case-sensitive"
-comparison in <cite>HTML</cite>. Strings that compare as <a>identical to</a> one another are not
-only sensitive to case variation (such as UPPER and lower case), but also to other code point
-encoding  choices, such as normalization form or the order of combining marks. Two strings that are
-visually or even canonically equivalent according to Unicode might still not be
+comparison in <cite>HTML</cite>. Strings that compare as <a for=string>identical to</a> one another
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -679,6 +679,20 @@ where <a>UTF-8 encode</a> comes into play.
 
 <hr>
 
+<p>A <a>string</a> <var>a</var> <dfn export>is</dfn> or is <dfn export>identical to</dfn> a
```
annevk, line 682: We should use `<dfn export for=string lt="is|identical to">is</dfn>` here and then use `<a for=string>is</a>` or `<a for=string>identical to</a>` when referring to it. The `for` attribute helps with disambiguation from other equality comparisons and if we allow multiple terms for the same definition we should list those in the definition so that all the links go to the right place.

infra.bs
```diff
@@ -679,6 +679,20 @@ where <a>UTF-8 encode</a> comes into play.
 
 <hr>
 
+<p>A <a>string</a> <var>a</var> <dfn export>is</dfn> or is <dfn export>identical to</dfn> a 
+<a>string</a> <var>b</var> if it consists of the same sequence of <a>code units</a>.
+
+<p>Except where otherwise stated, all string comparisons use this definition of string equality.
```
annevk, line 685:
```suggestion
<p>Except where otherwise stated, all string comparisons use <a for=string>is</a>.
```

infra.bs
```diff
@@ -679,6 +679,20 @@ where <a>UTF-8 encode</a> comes into play.
...
+
+<div class=note>
+ <p>This type of <a>string</a> comparison was formerly known as <dfn>case-sensitive</dfn>
```
annevk, line 688:
```suggestion
 <p>This type of <a>string</a> comparison was formerly known as a "case-sensitive" 
```

infra.bs
```diff
@@ -679,6 +679,20 @@ where <a>UTF-8 encode</a> comes into play.
...
 
+<p>A <a>string</a> <var>a</var> <dfn export>is</dfn> or is <dfn export>identical to</dfn> a 
+<a>string</a> <var>b</var> if it consists of the same sequence of <a>code units</a>.
+
+<p>Except where otherwise stated, all string comparisons use this definition of string equality.
+
+<div class=note>
+ <p>This type of <a>string</a> comparison was formerly known as <dfn>case-sensitive</dfn> 
+ comparison in HTML. Strings that compare as <a>identical to</a> one another are not only
+ sensitive case variation (such as UPPER and lower case), but also to other code point encoding
```
annevk, line 690:
```suggestion
 sensitive to case variation (such as UPPER and lower case), but also to other code point encoding 
```

infra.bs
aphillips: Addresssed in 774e2c1. Thanks for your help!!

infra.bs
```diff
@@ -679,6 +679,22 @@ where <a>UTF-8 encode</a> comes into play.
...
 <hr>
 
+<p>A <a>string</a> <var>a</var> <dfn export for=string lt="is|identical to">is</dfn> or is 
+<a for=string>identical to</a> a <a>string</a> <var>b</var> if it consists of the same sequence of 
+<a>code units</a>.
+
+<p>Except where otherwise stated, all string comparisons use <a for=string>is</a>.
+
+<div class=note>
+
```
annevk, line 689: This newline can be removed and we should use for=string in the "identical to" mentions below to reduce the amount of refactoring we need to do in the future. Looks good to me otherwise.

@domenic thoughts?

domenic: This should actually be a `<p class="note">`. I'm happy to push a fixup commit for that (and the other issue you mentioned).

infra.bs
```diff
@@ -679,6 +679,19 @@ where <a>UTF-8 encode</a> comes into play.
...
+<a>code units</a>.
+
+<p>Except where otherwise stated, all string comparisons use <a for=string>is</a>.
+
+<p class=note>This type of <a>string</a> comparison was formerly known as a "case-sensitive"
+comparison in <cite>HTML</cite>. Strings that compare as <a>identical to</a> one another are not
+only sensitive to case variation (such as UPPER and lower case), but also to other code point
+encoding  choices, such as normalization form or the order of combining marks. Two strings that are
+visually or even canonically equivalent according to Unicode might still not be
+<a for=string>identical to</a> each other. [[HTML]] [[UNICODE]]</p>
```
annevk, line 693:
```suggestion
<a for=string>identical to</a> each other. [[HTML]] [[UNICODE]]
```

infra.bs
aphillips: I tried to follow the local style :-). Thanks for the feedback.

annevk:
There's still one instance, I'll add it. And yeah, non-normative references seem okay to me.

annevk:
@aphillips one thing that's remaining here is that your membership of the Amazon WHATWG organization is not public. See https://help.github.com/en/github/setting-up-and-managing-your-github-user-account/publicizing-or-hiding-organization-membership for instructions.

aphillips:
@annevk Thanks. I was going to ask why that didn't get picked up. Should be fixed now.

annevk:
Oh, and we need to verify the organization. Will do.

annevk:
Thanks @aphillips!

PR #304, opened 2020-05-13
Handle JS values which JSON-stringify as undefined

### LINKED ISSUE #264: JSON serialization/deserialization to/from bytes doesn't handle undefined and 0 bytes
image

`JSON.stringify` can return undefined if called with things like `undefined` or `Symbol('mySymbol')`

This should result in 0 bytes (null) being the result of the algorithm. We can of course also just make it return an error and handle it elsewhere

deserialization should probably also assert that there is actually a non-zero length byte sequence

domenic:
Closes #264.

***
<a href="https://whatpr.org/infra/304.html" title="Last updated on May 13, 2020, 4:38 PM UTC (f478d10)">Preview</a> | <a href="https://whatpr.org/infra/304/a579d74...f478d10.html" title="Last updated on May 13, 2020, 4:38 PM UTC (f478d10)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -1357,6 +1357,12 @@ steps:
   <p class=note>Since no additional arguments are passed to <a>%JSONStringify%</a>, the resulting
   string will have no whitespace inserted.
 
+ <li>
+  <p>If <var>jsonString</var> is undefined, then throw a {{TypeError}}.
+
+  <p class="note">This can happen if <var>value</var> does not have a JSON representation, e.g., if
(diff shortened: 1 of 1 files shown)


domenic:
https://w3c.github.io/reporting/ passes an Array, so is OK. https://w3c.github.io/webauthn/#clientdatajson-serialization passes a "temporary copy" of... a dictionary, I think?

Those are all the consumers I could find on Google search.

Issue #230, opened 2018-12-08
List/contains definition unclear


littledan:
From https://infra.spec.whatwg.org/#list-contain:

> A list contains an item if it appears in the list. We can also denote this by saying that, for a list list and an index index, "list[index] exists". 

I'm confused about a couple points here:
- Is the equality algorithm somehow implied by "appears", or is it described somewhere else?
- How does the second sentence relate to the first?

(Context: I'm wondering about how to use an Infra set for the `[[PlatformBrand]]` (https://github.com/heycam/webidl/issues/97), and possibly also for private method brands (https://github.com/tc39/proposal-decorators/issues/180). It looks like I'd use this `contains` algorithm to check membership, right?)

annevk:
See also #106. I think we mostly want object identity here, except for primitive data types. Most specifications already assume those semantics when it comes to equality.

I think the second sentence doesn't really relate to the first that much and probably needs to be reworded. Maybe something like:

> For a list _list_ and an index _index_, _list_[_index_] exists, if _index_ is not out-of-bound for _list_.

annevk:
(I'd expect `[[PlatformBrand]]` to be a set, but yeah, you'd use contains.)

littledan:
Can sets contain things like Records? I'm not sure if those are considered to have object identity.

So the second sentence was just an idependent definition of "exists"?

littledan:
2. Is "appears" interpreted to reference the local definition of "exists"? It's unclear to me how.

annevk:
No, I think it's meant as looping over the list, comparing the input item with the items that result from looping, and returning true if they're "equal". Anyway, I agree with you that this could use some refactoring.

domenic:
Some definitions in Infra are fairly vacuous. E.g. contains delegates to an undefined "appears in" concept, iterate to "on each item in order", append and prepend to "add". The intention is roughly "you already know what this vague concept is, but here's the exact word you can use and reference".

We could probably redefine "contains" in terms of equals and iterate, instead of "appears". But I'm not sure how far we should pursue such a program. The endpoint of making these definitions rigorous is basically axiomatic set theory.

PR #305, opened 2020-05-15
Give a name to byte and code point's underlying number

annevk:
This would be useful for https://github.com/whatwg/url/pull/518 and also clarifies isomorphic decode/encode a bit.

***
<a href="https://whatpr.org/infra/305.html" title="Last updated on May 18, 2020, 4:32 AM UTC (5aa6822)">Preview</a> | <a href="https://whatpr.org/infra/305/0aad1a9...5aa6822.html" title="Last updated on May 18, 2020, 4:32 AM UTC (5aa6822)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -442,9 +442,9 @@ JavaScript <b>null</b> value. [[!ECMA-262]]
 
 <h3 id=bytes>Bytes</h3>
 
-<p>A <dfn export>byte</dfn> is a sequence of eight bits, represented as a double-digit hexadecimal
-number in the range 0x00 to 0xFF, inclusive. A <a>byte</a>'s <dfn for=byte>value</dfn> is its
-underlying number.
+<p>A <dfn export>byte</dfn> is a sequence of eight bits, represented as two
+<a>ASCII upper hex digits</a>, in the range 0x00 to 0xFF, inclusive. A <a>byte</a>'s
+<dfn for=byte>value</dfn> is its underlying number.
 
 <p class=example id=example-byte-value>0x40 is a <a>byte</a> whose <a for=byte>value</a> is 64.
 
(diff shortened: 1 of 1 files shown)


annevk:
We'd only have to do that in the return value, right? And that already has

> Return four bytes whose values

so it would only be matter of linking bytes and values.

I guess we could also say that a code point is always prefixed with U+... then and not typically.

domenic:
I see, so the variables named _byte1_ etc. would not actually be bytes, they'd be byte values? I guess that works.

> I guess we could also say that a code point is always prefixed with U+... then and not typically.

I've always found the idea that we allow 0x notation for code points weird. I haven't seen it in action but maybe it's done in MIME Sniff or Encoding.

annevk:
I hadn't looked in detail to Encoding before writing this, but looking at it now most of it seems to be written assuming this exists. 😊

I'll tidy up the language around code points as well though, to only allow them prefixed.

infra.bs
```diff
@@ -535,14 +538,17 @@ contains, in the range 0x61 (a) to 0x7A (z), inclusive, by 0x20.
...
+<a for="byte sequence">length</a> and whose <a>code points</a> have the same
+<a for="code point">values</a> as the <a for=byte>values</a> of <var>input</var>'s <a>bytes</a>, in
+the same order.
 
 
 <h3 id=code-points>Code points</h3>
 
 <p>A <dfn export lt="code point|character">code point</dfn> is a Unicode code point and is
-represented as a four-to-six digit hexadecimal number, typically prefixed with "U+".
+represented as "U+" followed by four-to-six <a>ASCII upper hex digits</a>, in the range U+0000 to
```
annevk, line 549: Is this too circular? Not really sure how else to do this.

aphillips: Maybe do it the other way around? Perhaps:

> A `code point` is a Unicode code point, whose `value` is an integer between 0 and 0x10FFFF inclusive and represented as the string `U+` followed by four to six `ASCII upper hex digits`.

annevk: I think that still has the same issue, as ASCII upper hex digits are defined using the U+ convention.

aphillips: Avoiding the circularity seems like it would lead to double-defining ASCII upper hex digits and reduce the overall clarity. I don't think people have a hard time with the U+ syntax so much as they have a difficult time with the difference between USVs/code points vs. code units (particularly the UTF-16 variety).

domenic: I don't think we can really avoid circularity in Infra without making it significantly more opaque. See previous discussions at https://github.com/whatwg/infra/issues/230#issuecomment-445774344

PR #310, opened 2020-06-01
Fix and symmetrize "less than" and "prefix"

### LINKED ISSUE #309: Byte sequences are "byte less than" their prefixes
Hi,

I noticed that the [byte less than](https://infra.spec.whatwg.org/#byte-less-than) operation on byte sequences produces unexpected results if one of the byte sequences [starts with](https://infra.spec.whatwg.org/#byte-sequence-starts-with) the other one (i.e. it has the other one as a prefix). For example, consider the byte sequences

_A_ = `` `Hello world` ``,
_B_ = `` `Hello` ``.

In determining whether or not _A_ is byte less than _B_, the algorithm proceeds as follows:

1. _B_ **does not** start with _A_, so move on to step 2.
2. _A_ **does** start with _B_ (`` `Hello world` `` starts with `` `Hello` ``), so return true.

This means that _A_ is byte less than _B_, or `` `Hello world` `` is byte less than `` `Hello` ``. This ordering is not consistent with the analogous ordering on code units, [code unit less than](https://infra.spec.whatwg.org/#code-unit-less-than). If we consider the corresponding strings

_sA_ = `"Hello world"`,
_sB_ = `"Hello"`,

we can see that _sB_ is a code unit prefix of _sA_, so _sA_ is **not** code unit less than _sB_ according to [the "code unit less than" algorithm](https://infra.spec.whatwg.org/#code-unit-less-than).

It seems to me that in [the "byte less than" algorithm](https://infra.spec.whatwg.org/#byte-less-than), the variables _a_ and _b_ should be swapped in step 1 and step 2. This would cause the algorithm to return false if _b_ is a "byte prefix" of _a_ (meaning that _a_ starts with _b_) and return true if _a_ is a "byte prefix" of _b_.

domenic:
We had a definition of (code unit) "prefix" for strings, but "starts
with" for byte sequences, which led to usage errors within the
corresponding "less than" algorithms.

This makes "prefix" the primary operation for both strings and byte
sequences, with dedicated `<dfn>`s for "starts with" for both. Then, it
fixes "less than" for byte sequences to use "prefix" instead of "starts
with", which makes the algorithm correct and closes #309.

Additionally, this fixes the code unit prefix algorithm to be correct
instead of backward. #259 accidentally made prefix behave like starts with, likely because it did not realize that we had different primary operations for strings vs. byte sequences.

@gabpiv, please take a look! Also let us know if you'd like to be acknowledged with a different name.

***

### :boom: Error: 400 Bad Request :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jun 2, 2020, 9:32 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fwhatwg%2Finfra%2F87650ca9a41e93100d05a4442c30793e670de98e%2Finfra.bs&force=1&md-status=LS-PR&md-Text-Macro=PR-NUMBER%20310)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20whatwg/infra%23310.)._
</details>


infra.bs
```diff
@@ -516,17 +516,20 @@ contains, in the range 0x61 (a) to 0x7A (z), inclusive, by 0x20.
  </li>
 </ol>
 
+<p>"<var>a</var> <dfn export for="byte sequence">starts with</dfn> <var>b</var> can be used as a
+synonym for "<var>b</var> is a <a for="byte sequence">prefix</a> of <var>a</var>".
```
annevk, line 520: Let's use `<dfn lt>` as we do elsewhere?

domenic: You mean, on the prefix definition? I think in this case we should have separate dfns, since they are literally the opposite of each other...

annevk: Oh wow, I missed that. Agreed.

annevk: What do you think about not using a and b but instead _prefix_ and _input_ or some such?

domenic: That's a great idea, will change.

domenic:
> Would it be possible to symmetrize these even more by changing

I thought about that. However, I think there's a bit of an important asymmetry here between strings and byte sequences. Byte sequences are unambiguously ... sequences of bytes. But strings could be interpreted either as sequences of code units, or as sequences of code points. So I think it's important to actually have "code unit" in the name of the concepts for strings.

You could define a separate set of concepts, "code point prefix" and "code point less than" for strings. We haven't done so because we actually want to discourage that. Given that, we could consider defining "prefix" as a synonym for "code point prefix". But, I think it's nicer to require callers to be explicit, so people can look at a sentence in some other spec and clearly understand it's a code unit prefix, not a code point prefix.

> I'd love to be listed as Gabriel Pivovarov if that's okay.

Will do!

gabpiv:
That is a very good point. I agree that it is better to have "code unit" in the name to be explicit! Thank you for your explanation.

aphillips:
@domenic noted:

>You could define a separate set of concepts, "code point prefix" and "code point less than" for strings. We haven't done so because we actually want to discourage people from comparing strings by code point in this way. (They should either use the very simple, deterministic "code unit" concepts, or they should use more complicated Unicode-aware algorithms.) 

The use of `code unit less than` has the side effect of ordering strings in UTF-16 order, that is, with all the supplementary characters in a kind of weird blob between U+D7FF and U+E000. This is fine because we only use it internally (never for display) and we don't really care about the logical ordering being... well... logical. Code unit order is easily computed and fast because practically speaking that's how the comparison is actually always implemented.

Code point order is equally deterministic and "very simple"--except that implementations always have to compute the code points (slow, involves allocation).

By "more complicated Unicode-aware algorithms", do you mean [Unicode collation](https://www.unicode.org/reports/tr10/) (i.e. linguistically aware sorting)?

I do think putting a note here explaining why everything is defined in terms of code units would be helpful: it would save having to repeat this reasoning.

> Given that, we could consider defining "prefix" as a synonym for "code point prefix". But, I think it's nicer to require callers to be explicit, so people can look at a sentence in some other spec and clearly understand it's a code unit prefix, not a code point prefix.

+1. The one weird side effect of this is that the "prefix" might not fall on a code point boundary (although this more or less never happens in practice).

annevk:
Yeah, it's always equal unless the string, when interpreted as containing code points, contains a (lone) surrogate. We should maybe consider asserting that the prefix does not contain a surrogate (or if we do have such callers, be very explicit about the implications).

aphillips:
Seems like Infra already addressed the need for a note (if memory--now jarred into action--serves, this was in response to my comment some time ago 😛):

> This matches the ordering used by JavaScript’s < operator, and its sort() method on an array of strings. This ordering compares the 16-bit code units in each string, producing a highly efficient, consistent, and deterministic sort order. The resulting ordering will not match any particular alphabet or lexicographic order, particularly for code points represented by a surrogate pair. [ECMA-262]

This definition is hardly used, but the point of it is to produce a deterministic ordering as efficiently as possible. The inputs will mostly be lists of well-formed strings, but if they contained isolated surrogates, code unit ordering them wouldn't be harmful that I can so. So I don't see adding a check for isolated surrogates.

annevk:
The case I was thinking of is where you want "code point prefix" and use "code unit prefix" because the former doesn't exist. But 1) that's a very questionable scenario to begin with and 2) that's a caller problem, not an algorithm problem.

PR #321, opened 2020-07-23
Specify algorithm declarations

### LINKED ISSUE #92: Sketch out prose for algorithm definitions.
It would be lovely if we could agree upon a standard way of describing algorithms in specs. For instance, it's helpful to understand expected inputs and outputs, but there's no commonly shared way of spelling those out. Some examples:

* WebAuthn has note blocks describing inputs: https://w3c.github.io/webauthn/#makeCredential, and describes outputs in prose.

* CSP describes both inputs and outputs in prose, usually in the form 'Given a request’s cryptographic nonce metadata (nonce) and a source list (source list), this algorithm returns "Matches" if the nonce matches one or more source expressions in the list, and "Does Not Match" otherwise:'.

* ECMAScript describes inputs but not outputs: "The abstract operation PerformEval with arguments x, evalRealm, strictCaller, and direct performs the following steps:"

And so on. It would be great if we could align this to enough of an extent that we could start building tooling support for the callsite as well.

domenic:
Closes #92.

***
<a href="https://whatpr.org/infra/321.html" title="Last updated on Aug 6, 2020, 5:35 AM UTC (146d27b)">Preview</a> | <a href="https://whatpr.org/infra/321/cdd0060...146d27b.html" title="Last updated on Aug 6, 2020, 5:35 AM UTC (146d27b)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -25,6 +25,13 @@ urlPrefix: https://tc39.github.io/ecma262/#; spec: ECMA-262;
         text: Type; url: sec-ecmascript-data-types-and-values
 </pre>
 
+<style>
+/* Used for normative exemplars of how to write algorithms, as distinct from .example */
+.exemplary-prose {
+  margin-left: 2em;
+}
+</style>
+
 
 <h2 id=goals class=no-num>Goals</h2>
 
@@ -239,10 +246,9 @@ resolution algorithm".
 
 <p>Declare algorithms by stating their name, parameters, and return type, in the following form:
 
-<blockquote>
-To <dfn ignore>[algorithm name]</dfn>, given a [type1] <var ignore>[parameter1]</var>, a [type2]
-<var ignore>[parameter2]</var>, &hellip;, perform the following steps. They return a [return type].
-</blockquote>
+<p class=exemplary-prose>To <dfn ignore>[algorithm name]</dfn>, given a [type1]
+<var ignore>[parameter1]</var>, a [type2] <var ignore>[parameter2]</var>, &hellip;, perform the
+following steps. They return a [return type].</p>
 
 <p>(For non-verb phrase algorithm names, use "To perform the
 <dfn ignore>[algorithm name]</dfn>&hellip;".)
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -226,6 +228,60 @@ as well as the single algorithm would. Therefore performance is best left as a f
...
+To <dfn ignore>[algorithm name]</dfn>, given a [type1] <var ignore>[parameter1]</var>, a [type2]
+<var ignore>[parameter2]</var>, &hellip;, perform the following steps. They return a [return type].
+</blockquote>
+
+<p>(For non-verb phrase algorithm names, use "To perform the
+<dfn ignore>[algorithm name]</dfn>&hellip;".)
+
+<p class=example id=example-algorithm-declaration>To <dfn ignore>parse an awesome format</dfn> given
+a [=byte sequence=] <var ignore>bytes</var>, perform the following steps. They return a
+[=string=] or null.
```
annevk, line 252: Typically we lead with null.

infra.bs
```diff
@@ -226,6 +228,60 @@ as well as the single algorithm would. Therefore performance is best left as a f
...
+
+<p class=example id=example-algorithm-declaration-names>Some algorithm names in the latter category
+include "attribute change steps", "internal module script graph fetching procedure", and "overload
+resolution algorithm".
+
+Declare algorithms by stating their name, parameters, and return type, in the following form:
+
+<blockquote>
+To <dfn ignore>[algorithm name]</dfn>, given a [type1] <var ignore>[parameter1]</var>, a [type2]
+<var ignore>[parameter2]</var>, &hellip;, perform the following steps. They return a [return type].
```
annevk, line 244: What's the ellipsis for?

infra.bs
```diff
@@ -198,7 +200,7 @@ leading spaces" or "return false") are to be interpreted with the meaning of the
  <p>it would be equivalent to the following:</p>
 
  <blockquote>
-  <p>To eat an orange:
+  <p>To <dfn ignore>eat an orange</dfn>:
```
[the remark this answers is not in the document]
domenic: It seems pretty unambiguous to me?

infra.bs
domenic: To indicate that you're not limited to only two parameters.

infra.bs
domenic: Hmm, I don't... I prefer to keep the more important type first.

infra.bs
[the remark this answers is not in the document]
annevk: Yeah, I guess it's okay. It's just that other places have a more direct connection with "steps" or "algorithm".

infra.bs
```diff
@@ -226,6 +228,60 @@ as well as the single algorithm would. Therefore performance is best left as a f
...
+<a lt="ASCII uppercase">ASCII uppercasing</a> the <a lt="isomorphic decode">ismorphic decoding</a>
+of <var ignore>bytes</var>.
+
+<p>Types should be included in algorithm declarations, but may be omitted if the parameter name is
+clear enough, or if they are otherwise clear from context. (For example, because the algorithm is
+a simple wrapper around another one.)
+
+<p class=example id=example-algorithm-declaration-no-types>To
+<dfn ignore>load a classic script</dfn> given <var>url</var>, return the result of performing the
+internal script-loading algorithm given <var>url</var> and "<code>classic</code>".
```
annevk, line 279: I typically use "with" instead of "given" when calling, but no strong opinion.

domenic:
Great catch. I think a class would be a bit nicer; let me give that a shot.

domenic:
Pushed; it seems reasonable. Note that I didn't change the blockquotes inside the examples, as they are "from another source" by my reading---albeit, a hypothetical source. I'm not solid either way though, so if you think they should also be changed I'm happy to do so.

PR #325, opened 2020-08-17
Specify optional and named parameters

### LINKED ISSUE #320: Named arguments and especially named optional arguments
This is a special case of #92.

Non-named boolean parameters are kind of a pain. Consider HTML's [obtain a worker/worklet agent](https://html.spec.whatwg.org/#obtaining-a-worker/worklet-agent). Call sites for it look like:

> Return the result of obtaining a worker/worklet agent given _outside settings_, _isShared_, and true.

You have to go refer to the algorithm to find out what the true is. (It's "can block".) This is kind of bad, and [is discouraged in programming languages](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap). Perhaps worse, you might think that the second argument is about being shared or not, but really what's happening is that we're passing in the value of _isShared_ for the _isTopLevel_ parameter to the original algorithm. The fact that the second argument is about top-levelness is totally lost.

I can imagine a variety of solutions:

0. Do nothing, and continue without named parameters.

1. Define a struct, e.g. containing "is top level" and "can block", and require callers to construct such a struct. Example: [navigation params](https://html.spec.whatwg.org/#navigation-params). This is not great; it's heavyweight for the caller, and feels like an abuse of structs.

2. Pass an Infra map. E.g. "Return the result of obtaining a worker/worklet agent given _outside settings_, «[ "is top level" → _isShared_, "can block" → true ]»." This is unprecedented, and feels like a bit of an abuse of maps, but maybe OK?

3. Define a new pattern for named parameters. For example:

	> To **obtain a worker/worklet agent**, given an environment settings object or null _outside settings_, a boolean **is top level**, and a boolean **can block**:
	> 1. Let _agentCluster_ be null.
	> 2. If [is top level] is true, then set _agentCluster_ to a new [agent cluster].
	> 3. ...
	>
	> and later someone uses it:
	>> ... return the result of [obtaining a worker/worklet agent] given _outside settings_, [is top level] set to _isShared_, and [can block] set to true.

	(here using bold to denote what would be `<dfn>`s, and brackets to denote what would be `<a>`s.)
	
	What I mildly dislike about this is that we no longer have variable names for some of our arguments; instead we refer to some of them with links. That's a bit weird. Maybe there's a fix?

What do people think of these?

----

This problem becomes more urgent when you want to have optional named parameters. HTML's [navigate](https://html.spec.whatwg.org/#navigate) already has one ("exceptions enabled flag"), and would benefit from at least two more, in its current formulation: replacement enabled (boolean), and navigation type ("form-submission" or "other"). Optionality seems important here; it would be very nice if we didn't have to pass "other" every time we had a non-form-submission navigation.

Of the above solutions, most of them work pretty well with optional named parameters. (1) could define defaults in its struct definition. (2) and (3) would involve some work in the algorithm to say something about if the map entry doesn't exist/the argument wasn't passed, then pick a default.

There is one other solution for this particular problem, though, which might be worth keeping in mind: define wrapper algorithms. So for example, we could define "internal navigate" which takes a non-optional navigation type, and then define "navigate" which calls "internal navigate" with navigation type of "other", and "form navigate" which calls "internal navigate" with navigation type of "form-submission". This might be a good thing, e.g. "form navigate" might never use replacement enabled or exceptions enabled. But in the general case it might be a bad thing; if all parameters are independent, then this leads to a combinatorial explosion of wrapper algorithms (imagine "form-replacement-with-exceptions navigate" vs. "form-replacement navigate" vs. "form navigate" vs. "form-with-exceptions navigate").

---

Anyway, it'd be good to straighten this out, as we start tidying up navigation in HTML.

domenic:
Closes #320.

***
<a href="https://whatpr.org/infra/325.html" title="Last updated on Oct 7, 2020, 6:28 PM UTC (a48ddfa)">Preview</a> | <a href="https://whatpr.org/infra/325/d91d797...a48ddfa.html" title="Last updated on Oct 7, 2020, 6:28 PM UTC (a48ddfa)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -239,7 +239,7 @@ as well as the single algorithm would. Therefore performance is best left as a f
 over.
 
 
-<h3 id=declaration>Declaration</h3>
+<h3 id=algorithm-declaration>Declaration</h3>
 
 <p>Algorithm names are usually verb phrases, but sometimes are given names that emphasize their
 standalone existence, so that standards and readers can refer to the algorithm more idiomatically.
@@ -255,7 +255,8 @@ resolution algorithm".
 following steps. They return a [return type].</p>
 
 <p>(For non-verb phrase algorithm names, use "To perform the
-<dfn ignore>[algorithm name]</dfn>&hellip;".)
+<dfn ignore>[algorithm name]</dfn>&hellip;". See also [[#algorithm-params]] for more complicated
+parameter-declaration forms.)
 
 <p class=example id=example-algorithm-declaration>To <dfn ignore>parse an awesome format</dfn> given
 a [=byte sequence=] <var ignore>bytes</var>, perform the following steps. They return a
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -282,8 +283,106 @@ a simple wrapper around another one.)
...
+<p class=note>Note how within the algorithm steps, the argument value is not linked to the parameter
+declaration; it remains just a variable reference. Linking to the parameter declaration is done only
+at the call sites.
+
+<p>Non-optional named parameters may also be used, using the same convention of marking them up as
+both variables and definitions, and linking to them from call sites. However, one should default to
+making non-optional parameters positional, for simplicity and uniformity. Non-optional named
+parameters should only be used when they significantly improve clarity, and even then it might be
+worth considering whether the algorithm would benefit from packaging up the related values into a
+<a>struct</a> instead.
```
annevk, line 385: What justifies allowing this from the get-go?

infra.bs
```diff
@@ -282,8 +283,106 @@ a simple wrapper around another one.)
...
+
+  <li><a href=#example-navigate-algo-named>Navigate</a> to <var ignore>resource</var> with
+  <a href=#example-navigate-algo-exceptionsEnabled><var ignore>exceptionsEnabled</var></a> set to
+  true.
+
+  <li><a href=#example-navigate-algo-named>Navigate</a> to <var ignore>resource</var> with
+  <a href=#example-navigate-algo-navigationType><var ignore>navigationType</var></a> set to
+  "<code>form submission</code>" and
+  <a href=#example-navigate-algo-exceptionsEnabled><var ignore>exceptionsEnabled</var></a> set to
+  true.
```
jakearchibald, line 372: If booleans default to false, should referencing them here set them to true? It seems to read well:

Navigate to resource with exceptionsEnabled.

infra.bs
```diff
@@ -282,8 +283,106 @@ a simple wrapper around another one.)
...
+ <p>Call sites to the previous example's algorithm would look like one of:
+
+ <ul class=brief>
+  <li><a href=#example-navigate-algo-positional>Navigate</a> to <var ignore>resource</var>.
+
+  <li><a href=#example-navigate-algo-positional>Navigate</a> to <var ignore>resource</var> with
+  "<code>form submission</code>".
+
+  <li><a href=#example-navigate-algo-positional>Navigate</a> to <var ignore>resource</var> with
+  "<code>form submission</code>" and true.
```
jakearchibald, line 328: Feels like we should call out positional booleans as bad practice.

infra.bs
annevk: Oh that's interesting. We should perhaps also allow/require omitting "(default false)" for booleans and basically have built-in support for that type, or is that too much?

infra.bs
annevk: If we really want to do that I guess that would further argue for special casing booleans in some ways.

infra.bs
domenic: I'd rather not special-case booleans in that way.

infra.bs
domenic: See the opening of #320.

infra.bs
domenic: Yeah, I dunno, it's OK-ish especially since it's all hypertext, but I agree generally it's not going to be a good time, and you probably want your booleans to be named.

infra.bs
annevk: I guess the question I have is why we'd still require the argument in those cases. (And to me it seems to again argue to special case booleans.)

domenic: I think the "obtain a worker/worklet agent" case, specifically, is a good example of why we'd want to require the argument. isTopLevel and canBlock don't have obvious defaults; making them optional seems like a bad idea to me.

infra.bs
domenic: In particular the benefit of "(default false)" is that it's obvious what the default is. And with "exceptionsEnabled set to true", it's obvious what the value of exceptionsEnabled is.

If you remove either of those, then the reader needs to know the Infra special-casing rules pretty intimately, otherwise they're left confused. "Why did this call site not pass exceptionsEnabled? Unlike navigationType, it has no default, so how is that possible?" "This call site said with exceptionsEnabled, but what is exceptionsEnabled actually set to? Are they trying to reference a nearby exceptionsEnabled variable's value?"

infra.bs
annevk: If all booleans always default to false I don't see why there would be confusion.

infra.bs
annevk: Yeah, I guess it's reasonable to require it for that reason, but we could require it to always be "(default false)" and forbid "(default true)".

infra.bs
domenic: I don't think it's correct to default those to false. Defaults should be used when the call sites overwhelmingly prefer the default, and you want to explicitly call out the non-default case. That's not true for "obtain a worker/worklet agent"; there's no obvious default to choose for those two booleans.

infra.bs
domenic: I'd be on board with that particular special-casing :).

infra.bs
annevk: It seems based on the discussion we should change this too, right? To only have examples with recommended practice.

domenic: Maybe. I'm unsure because it's part of the "narrative flow" of the section: it shows the algorithm evolving from unclear and inflexible optional positional arguments, to clear and flexible optional named arguments.

We could instead specifically explain that not only is this non-ideal because of the ordering restrictions, but also because of the presence of non-named boolean parameters?

annevk: Yeah that sounds good to me. I suspect that would work for @jyasskin too.

domenic: Done.

infra.bs
```diff
@@ -282,8 +283,117 @@ a simple wrapper around another one.)
...
+</div>
+
+<p class=note>Note how within the algorithm steps, the argument value is not linked to the parameter
+declaration; it remains just a variable reference. Linking to the parameter declaration is done only
+at the call sites.
+
+<p>Non-optional named parameters may also be used, using the same convention of marking them up as
+both variables and definitions, and linking to them from call sites. However, one should default to
+making non-optional parameters positional, for simplicity and uniformity. Non-optional named
+parameters should only be used when they significantly improve clarity, and even then it might be
```
jyasskin, line 388: FWIW, I think `Call foo with 3` and `Call foo with "navigate"` are not much better than `Call foo with false`, and Infra should encourage people to name those non-optional parameters. Perhaps:

```suggestion
<p>Non-optional parameters may also be named, using the same convention of marking them up as
both variables and definitions, and linking to them from call sites. Do this when it improves
clarity at call sites. It is also
```

infra.bs
jyasskin: Yep, thanks. Note that this example-of-things-we-don't-want-people-to-do winds up looking the same as all the other examples-of-things-to-do. Vanilla bikeshed supports a `heading="Bad Example"` attribute to make it a little clearer, but that attribute doesn't work with the [WHATWG style sheet](https://resources.whatwg.org/spec.css)'s `.example::before { content: 'Example'; }` rule.

domenic:
So, this has been sitting for a while. I'd like to get it landed. Outstanding feedback:

- @jyasskin's preference toward more named parameters, or at least against the current PR's avoidance of them, expressed in comments immediately above. I'm on the fence here. I like the idea of Infra giving guidance, but since this seems controversial, maybe Infra should be more neutral, and leave this up to editor preference.

- @jakearchibald's feedback in https://github.com/whatwg/infra/issues/320#issuecomment-692785334 that we shouldn't use `<var>` at call sites. Although I recognize the logic of his position, it feels a bit icky to have camelCased things that aren't `<var>`, and to have the typography at the call site mismatch the definition site. So I'm weakly for the PR as-is.

@annevk, any thoughts? And if anyone wants to chime with their perspective on the above, feel free to do so.

annevk:
I think encouraging naming for optional arguments is reasonable, but I'm okay with not endorsing anything for now, especially as there is no tooling support either.

We could use `<i>` at the call site instead of `<var>`.

jakearchibald:
With `<var>` you have to work around it not-really-being-a-var with `<var ignore>`. With `<i>` you don't need to work around it, so it gets my vote.

domenic:
Alright, this might be ready to go then!

jyasskin (approved):
Thanks!

annevk (approved):

PR #338, opened 2020-10-01
More JSON algorithms

### LINKED ISSUE #336: Serialize Infra values to JSON bytes
We have https://infra.spec.whatwg.org/#serialize-json-to-bytes which takes JS values, but nothing that takes Infra values. This would be convenient in https://w3c.github.io/webdriver-bidi/, where we use Infra sets, and then want to serialize those and send as a WebSocket message.

domenic:
The first commit here fixes #336. The second is a followup to help the section be more uniform and useful. They're likely best reviewed separately.

***
<a href="https://whatpr.org/infra/338.html" title="Last updated on Oct 5, 2020, 4:03 PM UTC (821937b)">Preview</a> | <a href="https://whatpr.org/infra/338/782c714...821937b.html" title="Last updated on Oct 5, 2020, 4:03 PM UTC (821937b)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -1427,55 +1427,81 @@ as 200/`<code>OK</code>`.
 <p class=note>The conventions used in the algorithms in this section are those of the JavaScript
 specification. [[!ECMA-262]]
 
-<p>To <dfn export>parse JSON from bytes</dfn> given <var>bytes</var>, run these steps:
+<p>To <dfn export lt="parse a JSON string to a JavaScript value|parsing a JSON string to a JavaScript value">parse a JSON string to a JavaScript value</dfn>,
+given a <a>string</a> |string|:
 
 <ol>
- <li><p>Let <var>jsonText</var> be the result of running <a>UTF-8 decode</a> on <var>bytes</var>.
- [[!ENCODING]]
-
- <li><p>Return ? <a abstract-op>Call</a>(<a>%JSON.parse%</a>, undefined, « <var>jsonText</var> »).
+ <li><p>Return ? [$Call$](<a>%JSON.parse%</a>, undefined, « |string| »).
 </ol>
 
-<p>To <dfn export>serialize JSON to bytes</dfn> a given JavaScript value <var>value</var>, run these
-steps:
+<p>To <dfn export lt="parse JSON bytes to a JavaScript value|parsing JSON bytes to a JavaScript value|parse JSON from bytes">parse JSON bytes to a JavaScript value</dfn>,
+given a <a>byte sequence</a> |bytes|:
+
+<ol>
+ <li><p>Let |string| be the result of running <a>UTF-8 decode</a> on |bytes|. [[!ENCODING]]
+
+ <li><p>Return the result of <a>parsing a JSON string to a JavaScript value</a> given |string|.
+</ol>
+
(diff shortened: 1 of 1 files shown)


foolip (approved):

infra.bs
```diff
@@ -1427,55 +1427,81 @@ as 200/`<code>OK</code>`.
...
- [[!ENCODING]]
+ <li><p>Return ? [$Call$](<a>%JSON.parse%</a>, undefined, « |string| »).
+</ol>
+
+<p>To <dfn export lt="parse JSON bytes to a JavaScript value|parsing JSON bytes to a JavaScript value|parse JSON from bytes">parse JSON bytes to a JavaScript value</dfn>,
+given a <a>byte sequence</a> |bytes|:
 
- <li><p>Return ? <a abstract-op>Call</a>(<a>%JSON.parse%</a>, undefined, « <var>jsonText</var> »).
+<ol>
+ <li><p>Let |string| be the result of running <a>UTF-8 decode</a> on |bytes|. [[!ENCODING]]
```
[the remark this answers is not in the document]
domenic: Yeah, it looks like by default, the error mode is "replacement". You can use "UTF-8 decode without BOM or fail" to get actual errors.

infra.bs
```diff
@@ -1522,6 +1551,82 @@ given a JavaScript value <var>jsValue</var>:
...
+ |value|.
+
+ <li><p>Return the result of running <a>UTF-8 encode</a> on |string|. [[!ENCODING]]
+</ol>
+
+<p>To <dfn lt="convert an Infra value to a JSON-compatible JavaScript value|converting an Infra value to a JSON-compatible JavaScript value">convert an Infra value to a JSON-compatible JavaScript value</dfn>,
+given |value|:
+
+<ol>
+ <li><p>If |value| is a <a>string</a>, <a>boolean</a>, number, or null, return |value|.
```
andreubotella, line 1584: The remaining primitives are bytes and code points. Bytes, although defined in Infra as "a sequence of eight bits" (\**shudders*\*), are basically used everywhere as numbers, so they should be serialized as such. But should code points be serialized as a string?

andreubotella (commented):
I think it'd be better to have all Infra types be serializable, even if they don't round-trip.

domenic:
I think it's better not to stretch things so that bytes/code points/byte sequences/structs/maps-with-non-string-keys become serializable somehow. The goal isn't to provide a mapping between Infra and JSON. It's to allow you to set up specially-crafted Infra values, and turn them into JSON.

I'd be happy to revisit that if we had concrete use cases, but I suspect that people who are creating Infra values with an eye toward JSON serialization will just not use the unserializable types.

foolip:
For WebDriver BiDi we will indeed be crafting Infra values specifically to pass to this algorithm, we won't be passing in values that we also use for some other internal purpose.

andreubotella:
Fair enough. Since stacks, queues and sets are subsets of list, they could still be serialized, but that's probably not worth disallowing.

I've deleted the comments relating to serialization of further types above, leaving the typo fix.

infra.bs
```diff
@@ -1522,6 +1525,82 @@ given a JavaScript value <var>jsValue</var>:
...
```
annevk, line 1584: then return

PR #342, opened 2020-10-07
Editorial: avoid saying isolated surrogates

### LINKED ISSUE #341: Isolated Surrogates
In https://infra.spec.whatwg.org/#strings you write:

> Note: The replaced surrogates are always isolated surrogates, since the process of interpreting the string as containing code points will have converted surrogate pairs into scalar values. ...

This is not correct.  A counter example is a string composed of two surrogates:

U+D800 U+D800

They do not form a valid surrogate pair, as they are both from the low-surrogate range. Unicode demands the first being a low surrogate, the second a high-surrogate.  This is not the case here, they can never represent a Unicode codepoint.  In such a case you most likely want both be replaced with U+FFFD REPLACEMENT CHARACTER.  Obviously, the surrogates are NOT isolated.

Wolf Lammen

annevk:
Fixes #341.

@aphillips want to review?

***
<a href="https://whatpr.org/infra/342.html" title="Last updated on Jan 15, 2021, 7:48 AM UTC (074846c)">Preview</a> | <a href="https://whatpr.org/infra/342/d91d797...074846c.html" title="Last updated on Jan 15, 2021, 7:48 AM UTC (074846c)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -712,8 +712,8 @@ refers exclusively to how the Unicode Standard defines it for Unicode 16-bit str
 defined in <a>The String Type</a> section of the JavaScript specification. [[!ECMA-262]]
 
 <p class=note>This conversion process converts surrogate pairs into their corresponding
-<a>scalar value</a> and maps isolated surrogates to their corresponding <a>code point</a>, leaving
-them effectively as-is.
+<a>scalar value</a> and maps any remaining surrogates to their corresponding <a>code point</a>,
+leaving them effectively as-is.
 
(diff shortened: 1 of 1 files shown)


aphillips (approved):
Looks good to me!

rmisev:
Isn't it better to use the term "Unpaired surrogate" here? It is from: https://unicode.org/faq/utf_bom.html#utf16-7

domenic (approved):

annevk:
I looked at the various references again in the original issue and that term isn't formally defined either, just like isolated, so I'm going with the clarification I made here.

PR #386, opened 2021-07-13
Define exclusive ranges and list indices

domenic:
Also add "keys" and "values" as linking aliases for "get the keys"/"get the values" on maps.

This will help generalize https://github.com/whatwg/html/pull/6738

***
<a href="https://whatpr.org/infra/386.html" title="Last updated on Jul 13, 2021, 5:31 PM UTC (aa4ba22)">Preview</a> | <a href="https://whatpr.org/infra/386/b372703...aa4ba22.html" title="Last updated on Jul 13, 2021, 5:31 PM UTC (aa4ba22)">Diff</a>


annevk (approved):
We have a single instance of indexes in the HTML Standard we probably want to replace.

PR #391, opened 2021-08-12
Add code unit/point substring definitions

### LINKED ISSUE #152: Investigate "substring" usage in URL and maybe add it
And elsewhere if it's used. Seems good to ground it somehow.

domenic:
Closes #152.

@wanderview would use this in URLPattern.

***
<a href="https://whatpr.org/infra/391.html" title="Last updated on Aug 12, 2021, 7:51 PM UTC (36e045a)">Preview</a> | <a href="https://whatpr.org/infra/391/d69ac66...36e045a.html" title="Last updated on Aug 12, 2021, 7:51 PM UTC (36e045a)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -941,6 +941,69 @@ ordering will not match any particular alphabet or lexicographic order, particul
 
 <hr>
 
+<p>The <dfn export>code unit substring</dfn> from index <var>start</var> with length
+<var>length</var> within a <a>string</a> <var>string</var> is determined as follows:
+
+<ol>
+ <li><p><a>Assert</a>: <var>start</var> and <var>length</var> are greater than 0.</p></li>
+
+ <li><p><a>Assert</a>: <var>start</var> + <var>length</var> is less than or equal to
+ <var>string</var>'s <a for=string>length</a>.</p></li>
+
+ <li><p>Let <var>result</var> be the empty string.</p></li>
+
+ <li><p><a for="set">For each</a> <var>i</var> in <a lt="the exclusive range">the range</a> from
+ <var>start</var> to <var>start</var> + <var>length</var>, exclusive: append the <var>i</var>th
+ <a>code unit</a> of <var>string</var> to <var>result</var>.</p></li>
+
+ <li><p>Return <var>result</var>.</li>
+</ol>
+
+<p>The <dfn export lt="code unit substring by indices">code unit substring</dfn> from indices
+<var>start</var> to <var>end</var> inclusive within a <a>string</a> <var>string</var> is the <a>code
+unit substring</a> from <var>start</var> with length <var>end</var> &minus; <var>start</var> within
+<var>string</var>.
+
+<p class="example" id="example-code-unit-substring">The <a>code unit substring</a> from index 1 with
+length 3 within "<code>Hello world</code>" is "<code>ell</code>". This can also be expressed as the
+<a lt="code unit substring by indices">code unit substring</a> from indices 1 to 4 inclusive.
+
+<p>The <dfn export>code point substring</dfn> within a <a>string</a> <var>string</var> from index
+<var>start</var> with length <var>length</var> is determined as follows:
+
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -941,6 +941,69 @@ ordering will not match any particular alphabet or lexicographic order, particul
...
 <hr>
 
+<p>The <dfn export>code unit substring</dfn> from index <var>start</var> with length
+<var>length</var> within a <a>string</a> <var>string</var> is determined as follows:
+
+<ol>
+ <li><p><a>Assert</a>: <var>start</var> and <var>length</var> are greater than 0.</p></li>
+
+ <li><p><a>Assert</a>: <var>start</var> + <var>length</var> is less than or equal to
+ <var>string</var>'s <a for=string>length</a>.</p></li>
```
annevk, line 951: I wonder if this is the most useful interface for all callers as it means they have to check the length of each string as well and do a bunch of branching around it.

infra.bs
```diff
@@ -941,6 +941,69 @@ ordering will not match any particular alphabet or lexicographic order, particul
...
+</ol>
+
+<p>The <dfn export lt="code unit substring by indices">code unit substring</dfn> from indices
+<var>start</var> to <var>end</var> inclusive within a <a>string</a> <var>string</var> is the <a>code
+unit substring</a> from <var>start</var> with length <var>end</var> &minus; <var>start</var> within
+<var>string</var>.
+
+<p class="example" id="example-code-unit-substring">The <a>code unit substring</a> from index 1 with
+length 3 within "<code>Hello world</code>" is "<code>ell</code>". This can also be expressed as the
+<a lt="code unit substring by indices">code unit substring</a> from indices 1 to 4 inclusive.
```
annevk, line 969:
```suggestion
<a lt="code unit substring by indices">code unit substring</a> from indices 1 to 4, inclusive.
```

infra.bs
```diff
@@ -941,6 +941,69 @@ ordering will not match any particular alphabet or lexicographic order, particul
...
+
+<p>The <dfn export lt="code point substring by indices">code point substring</dfn> from indices
+<var>start</var> to <var>end</var> inclusive within a <a>string</a> <var>string</var> is the <a>code
+point substring</a> within <var>string</var> from <var>start</var> with length <var>end</var>
+&minus; <var>start</var>.
+
+<div class="example" id="example-code-unit-vs-point-substring">
+ <p>Generally, <a>code unit substring</a> is used when given developer-supplied indices or lengths,
+ since that is how string indexing works in JavaScript. See, for example, the methods of the
+ {{CharacterData}} class. [[DOM]]
```
annevk, line 997: I wonder if we would design DOM this way if starting out fresh, but okay.

annevk (approved):
This seems fine to land as-is modulo the comma nit, but we might want to tweak the assertions going forward (at which point we would have to define what to return in those situations).

infra.bs
domenic: Yeah, I saw that URL seems to want to go to the end of the string. We could make length optional in a followup.

infra.bs
domenic: I didn't do this because then you need a lot more commas. In particular you can no longer say "Concatenate the substring from indices X to Y inclusive within Z with W". Instead you need "Concatenate the substring from indices Y to Z, inclusive, within Z, with W" which feels a good bit more awkward.

infra.bs
domenic: I mean, I think we would just have no methods or `length` proeprty at all, and just have people modify or read directly from `.data` :)

infra.bs
annevk: I see, I'm not sure I would have a good intuition as to when to include and when to omit them, as thus far I think I have consistently used a comma before inclusive (and sometimes after, if there's not a period). It does seem reasonable to omit it in a longer sentence like that though.

PR #401, opened 2021-09-09
Add substrings to the end of the string

domenic:
We have at least two consumers: https://github.com/WICG/urlpattern/pull/134 and various places in URL.

***
<a href="https://whatpr.org/infra/401.html" title="Last updated on Sep 9, 2021, 10:23 PM UTC (070ea39)">Preview</a> | <a href="https://whatpr.org/infra/401/0333d3d...070ea39.html" title="Last updated on Sep 9, 2021, 10:23 PM UTC (070ea39)">Diff</a>


annevk (approved):
I was initially thinking this should go below the examples, but looking at the text it seems reasonable.

PR #403, opened 2021-09-16
Explain how browser profiles map to user agents.

### LINKED ISSUE #400: Deal with profiles in definition/use of user agent
Several specifications have text like

> [A user agent holds a storage shed](https://storage.spec.whatwg.org/#storage-sheds)

> [A user agent holds a browsing context group set](https://html.spec.whatwg.org/multipage/browsers.html#browsing-context-group-set)

where we probably want one per browser profile / incognito or guest session / Firefox container / etc, instead of one per installation. (and definitely not one per [implementation](https://infra.spec.whatwg.org/#user-agent)!) We should do something to make sure that the text in specifications is clear about this.

It's not obvious whether we should redefine "user agent" to make the existing text correct, or introduce a new term and update all the uses.

jyasskin:
Fixes #400.

FYI @dbaron @mnot

***
<a href="https://whatpr.org/infra/403.html#terminology" title="Last updated on Sep 20, 2021, 4:09 PM UTC (a984d3d)">Preview</a> | <a href="https://whatpr.org/infra/403/ff425b4...a984d3d.html" title="Last updated on Sep 20, 2021, 4:09 PM UTC (a984d3d)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -120,8 +120,8 @@ A person can use many different [=user agents=] in their day-to-day life, includ
 or the implementation's private browsing mode.
 
 <p>If something is said to be <dfn export>implementation-defined</dfn>, the particulars of what is
-said to be <a>implementation-defined</a> are up to the <a>user agent</a>. In the absence of such
-language, the reverse holds: <a>user agents</a> have to follow the rules laid out in documents using
+said to be <a>implementation-defined</a> are up to the <a>implementation</a>. In the absence of such
+language, the reverse holds: <a>implementations</a> have to follow the rules laid out in documents using
 this standard.
 
 <p class="example" id=example-implementation-defined>Insert U+000A (LF) code points into
--- a/infra.bs
+++ b/infra.bs
@@ -111,11 +111,13 @@ out as being exclusive (with "but not both").
 
 <hr>
 
-<p>A <dfn export lt="user agent|implementation">user agent</dfn> is any software that acts on behalf
+<p>A <dfn export>user agent</dfn> is any software entity that acts on behalf
 of a user, for example by retrieving and rendering web content and facilitating end user interaction
-with it. In specifications using the Infra Standard, the user agent is generally the client software
-that implements the specification. <a>Implementation</a> can be used as a synonym for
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -111,11 +111,13 @@ out as being exclusive (with "but not both").
...
-with it. In specifications using the Infra Standard, the user agent is generally the client software
-that implements the specification. <a>Implementation</a> can be used as a synonym for
-<a>user agent</a>.
+with it. In specifications using the Infra Standard, the user agent is generally an instance of the client software
+that implements the specification. The client software itself is known as an <dfn export>implementation</dfn>.
+A person can use many different [=user agents=] in their day-to-day life, including by configuring an
+[=implementation=] to act as several [=user agents=] at once, for example by using multiple profiles
+or the implementation's private browsing mode.
 
 <p>If something is said to be <dfn export>implementation-defined</dfn>, the particulars of what is
```
domenic, line 122: Should this be changed to be "up to the implementation" and otherwise talk about implementations?

domenic (commented):
Nice improvement in general, just one question.

annevk:
I'm not sure if we want to make this so concrete. I'd rather say that a user can use different instances of a user agent or implementation, with different configurations. Making a formal distinction between implementation and user agent would probably require a lot of work downstream (and here too, as domenic points out).

jyasskin (commented):
Re @annevk, if different profiles are different "instances" of a single "user agent", then text like "A user agent holds a storage shed" is incorrect: it should be "A user agent _instance_ holds a storage shed". I'm fine with us choosing that change instead, but it's also a lot of work downstream.

jyasskin:
We could also [willfully violate](https://infra.spec.whatwg.org/#willful-violation) English and keep "Implementation can be used as a synonym for user agent.", with the note that willful violations need?

fallingicarus (approved):

fallingicarus (approved):
Approve

jyasskin:
Was there anything else to do before merging this?

PR #446, opened 2022-05-02
Add code unit suffix (ends with for strings)

### LINKED ISSUE #414: String "ends with"
https://fetch.spec.whatwg.org/#data-url-processor could use this. Though currently it passes starts with and ends with a code point. Not sure if there are other places that might need this?

annevk:
Closes #414.

(FWIW, I had an initial version that cached length before the loop in a variable, but I think it reads better without doing that.)

***
<a href="https://whatpr.org/infra/446.html" title="Last updated on May 2, 2022, 2:55 PM UTC (85c71c2)">Preview</a> | <a href="https://whatpr.org/infra/446/c178025...85c71c2.html" title="Last updated on May 2, 2022, 2:55 PM UTC (85c71c2)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -928,16 +928,18 @@ literal, we can use plainer language: <var>userInput</var> <a for="string">start
    <li><p>Let <var>potentialSuffixIndex</var> be <var>potentialSuffix</var>'s
    <a for=string>length</a> &minus; <var>i</var>.
 
+   <li><p>Let <var>inputIndex</var> be <var>input</var>'s <a for=string>length</a> &minus;
+   <var>i</var>.
+
    <li><p>If <var>potentialSuffixIndex</var> is less than 0, then return true.
+
+   <li><p>If <var>inputIndex</var> is less than 0, then return false.
 
    <li><p>Let <var>potentialSuffixCodeUnit</var> be the <var>potentialSuffixIndex</var>th
    <a>code unit</a> of <var>potentialSuffix</var>.
 
-   <li><p>Let <var>inputIndex</var> be <var>input</var>'s <a for=string>length</a> &minus;
-   <var>i</var>.
-
    <li><p>Let <var>inputCodeUnit</var> be the <var>inputIndex</var>th <a>code unit</a> of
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -915,6 +915,48 @@ as a synonym for "<var>potentialPrefix</var> is a <a>code unit prefix</a> of <va
...
+   <li><p>If <var>potentialSuffixIndex</var> is less than 0, then return true.
+
+   <li><p>Let <var>potentialSuffixCodeUnit</var> be the <var>potentialSuffixIndex</var>th
+   <a>code unit</a> of <var>potentialSuffix</var>.
+
+   <li><p>Let <var>inputIndex</var> be <var>input</var>'s <a for=string>length</a> &minus;
+   <var>i</var>.
+
+   <li><p>Let <var>inputCodeUnit</var> be the <var>inputIndex</var>th <a>code unit</a> of
+   <var>input</var> if <var>i</var> is greater than or equal to 0; otherwise null.
```
domenic, line 940: Is _i_ supposed to be _inputIndex_ here?

domenic (commented):
Optional feedback: it took me a while to realize the interacting roles of _i_, _potentialSuffixIndeX_, and _inputIndex_. I might do something like this:

```
1. Let i be 1.
2. While true:
  1. Let potentialSuffixIndex = potentialSuffix.length - i.
  2. Let inputIndex = input.length - i.
  3. If potentialSuffixIndex < 0, then return true.
  4. If inputIndex < 0, then return false.
  5. Let potentialSuffixCodeUnit = potentialSuffix[potentialSuffixIndex].
  6. Let inputCodeUnit = input[inputIndex].
  7. Return false if potentialSuffixCodeUnit != inputCodeUnit.
  8. Set i to i + 1.
```

However I can easily see how what you have here is closer to "code unit prefix", and so then you could get carried away trying to make them both use something like the above... so, very optional feedback.

annevk:
I'll make a follow-up PR for aligning the two "prefix" algorithms. Not too much effort.

domenic (approved):

Issue #449, opened 2022-05-05
Defer to CSS Text for definition of “character”


fantasai:
The Infra spec aliases its definition for “code point” to “character”. However, “character” is a very ambiguous term, see https://www.w3.org/TR/css-text-3/#characters, and should not be treated as an alias of “code point”. This isn't to say that specs shouldn't ever use it, if it's clear from context which meaning is meant, and it's fine to leave in the sentence that says

> [Code points](https://infra.spec.whatwg.org/#code-point) are sometimes referred to as [characters](https://infra.spec.whatwg.org/#code-point) and in certain contexts are prefixed with "0x" rather than "U+".

But I think it's best if Infra does not define and export the term “character” as an alias of “code point”.

domenic:
I disagree. I think if anyone is using the term "character" on the web platform, they should be linking to this clear definition as a synonym of "code point".

fantasai:
I think my point is that people shouldn't really be using the term “character” when defining behavior in Web platform specs, so Infra shouldn't be exporting the term and encouraging its use.

https://www.w3.org/TR/charmod/#sec-PerceptionsOutro

aphillips:
I'll pile on here.

On the one hand, I like WHATWG's general preference for using normal words in preference to jargon and I don't think that the definition of `character` in Infra is "wrong". 

However, spec developers don't understand the term that well: my worry is that they'll pick up the definition and use it imprecisely. When I18N does reviews, we often find text talking about "characters" but the assumption behind it is slipping around between bytes, code units (which obviously can _be_ bytes), code points, and graphemes. By getting specs to be specific, we avoid headaches later. Note that I18N WG made a conscious choice not to export `character` in the [I18N Glossary](https://www.w3.org/TR/i18n-glossary/#c).

domenic:
Sure, I think people should prefer more precise terms. I don't think that really has bearing on whether it's exported. And it certainly doesn't indicate we should defer to CSS Text (which also exports a definition---and IMO a worse one for the purposes of other specs) instead.

fantasai:
It's not so much that I want specs to refer to CSS Text, it's that I want them to recognize that it's an ambiguous term that has multiple definitions, and needs to be made more specific if it's being used in a spec.

> Sure, I think people should prefer more precise terms. I don't think that really has bearing on whether it's exported. 

[...]

PR #450, opened 2022-05-05
Don't export character as an alias of code point

fantasai:
“Character” is an ambiguous term and is used in multiple senses in the Web platform. Don't alias it to only one of them. #449

***
<a href="https://whatpr.org/infra/450.html" title="Last updated on May 5, 2022, 2:44 AM UTC (fe1656c)">Preview</a> | <a href="https://whatpr.org/infra/450/3deb8ac...fe1656c.html" title="Last updated on May 5, 2022, 2:44 AM UTC (fe1656c)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -728,7 +728,7 @@ the same order.
 
 <h3 id=code-points>Code points</h3>
 
-<p>A <dfn export lt="code point|character">code point</dfn> is a Unicode code point and is
+<p>A <dfn export>code point</dfn> is a Unicode code point and is
 represented as "U+" followed by four-to-six <a>ASCII upper hex digits</a>, in the range U+0000 to
 U+10FFFF, inclusive. A <a>code point</a>'s <dfn export for="code point">value</dfn> is its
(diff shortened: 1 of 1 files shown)


fantasai:
OK, so:

- The build claims to be failing because “character” can be referencing that term's definition in two different specs, css-text or infra, but the pull request is in fact removing the second definition (afaict; that *is* the point of this PR).
- I don't know in what capacity I'm supposed to sign the participation agreement, it's not clear that any of the categories applies to me.

Feel free to close the PR and make the edits on your own, I don't know how to untangle any of this.

domenic:
> * I don't know in what capacity I'm supposed to sign the participation agreement, it's not clear that any of the categories applies to me.

From what I understand you work as a contractor in the field of web technologies so should be covered under one or more of the entities you're contracting for.

> Feel free to close the PR and make the edits on your own, I don't know how to untangle any of this.

Given that the discussion in #449 is still ongoing, I'll close for now.

PR #451, opened 2022-05-05
Precisely specify iteration details. Fixes #396

### LINKED ISSUE #396: "For each" is not clear what happens if the list is modified while iterating
Apparently I do this sometimes: https://wicg.github.io/app-history/#inform-app-history-about-canceling-traversals step 2 calls into https://wicg.github.io/app-history/#finalize-with-an-aborted-navigation-error which in step 10.2 will remove the item from the list.

Options:

1. Disallow this and force the spec to make a copy of the list before iterating
2. Create a copy automatically when the iteration starts
3. Do something more complicated where deletions are handled transparently. I think this would look like:
   - If you delete something from the list before the currently-being-iterated value, that's OK. We ensure that it's you move on to the next value correctly (and don't e.g. skip forward 2 as an index-based approach would).
   - If you delete the currently-being-iterated value, that's OK. We ensure that you move on to the next value correctly (and don't e.g. skip forward 2 as an index-based approach would).
   - If you delete something after the currently-being-iterated value, that's OK. We ensure that it won't be seen in the rest of the iteration.

Thoughts welcome!

tabatkins:
Added a "sequence iteration" concept, and invoked it explicitly for list and map iteration.

Sequence iteration is defined in the same terms that ES does (see <https://github.com/tc39/ecma262/pull/2766> where ES is making itself more consistent in this regard): you maintain an internal index, incrementing it each step, and checking it against the length at each iteration. If the sequence changes size or order while executing its steps, that's fine, the next iteration will just see whatever the i+1th value of the sequence now is.

Happy to invoke this in a different way; what I have just seemed like a short and straightforward way to do so.

***
<a href="https://whatpr.org/infra/451.html" title="Last updated on Dec 13, 2022, 10:18 PM UTC (cf68dbb)">Preview</a> | <a href="https://whatpr.org/infra/451/3deb8ac...cf68dbb.html" title="Last updated on Dec 13, 2022, 10:18 PM UTC (cf68dbb)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -1319,6 +1319,12 @@ out-of-bounds, except when used with <a for=list>exists</a>.
 "<code>b</code>", "<code>c</code>", "<code>a</code>" ». Then |example|[1] is the <a>string</a>
 "<code>b</code>".
 
+<p>Lists can also have <dfn export for=list,stack,queue,set lt=hole>holes</dfn>, which mark the location of a deleted value. Holes are not exposed to <em>any</em> algorithm unless explicitly called out; in all other cases, holes are treated as if they did not exist and are automatically skipped over. There is no literal syntax for holes.
+
+<p class=example id=example-hole-skipping>For example, if the list « <code>0</code>, <code>1</code>, <code>2</code> » has the item at index 1 [=list/removed=], it leaves behind a hole in its position. Subsequent requests for the item at index 1 will now return <code>2</code>, as index counting skips over holes unless otherwise specified; similarly, the list will report its [=list/size=] as 2 rather than 3.
+
+Note: Holes exist to make [=list/iteration=] stable and predictable when the list can be mutated during the iteration, matching ECMAScript behavior. If the [=list=] is read-only or not currently being iterated by anything, [=list/holes=] can be omitted or removed.
+
 <hr>
 
 <p>To <dfn export for=list>append</dfn> to a <a>list</a> that is not an <a>ordered set</a> is to
@@ -1356,7 +1362,7 @@ index is to add the given item to the list between the given index &minus; 1 and
 the given index is 0, then <a for=list>prepend</a> the given item to the list.
 
 <p>To <dfn export for=list,set>remove</dfn> zero or more <a for=list>items</a> from a <a>list</a> is
-to remove all items from the list that match a given condition, or do nothing if none do.
+to remove all items from the list that match a given condition, replacing each with a [=list/hole=], or do nothing if none do.
 
 <div class=example id=example-list-remove>
  <p><a for=list>Removing</a> |x| from the <a>list</a> « |x|, |y|, |z|, |x| » is to remove all
@@ -1369,7 +1375,7 @@ to remove all items from the list that match a given condition, or do nothing if
 </div>
 
 <p>To <dfn export for=list,stack,queue,set>empty</dfn> a <a>list</a> is to <a for=list>remove</a>
-all of its <a for=list>items</a>.
+all of its <a for=list>items</a>, replacing all of them with [=list/holes=].
 
 <p>A <a>list</a> <dfn export for=list,stack,queue,set lt=contain|exist>contains</dfn> an
 <a for=list>item</a> if it appears in the list. We can also denote this by saying that, for a
@@ -1395,18 +1401,20 @@ using |list| and |steps|. This can also be (and usually is) written as "[=list/F
     <dfn>List iteration</dfn> over a [=list=] |list|, given some per-item |steps|, means to:
 
     1. Let |index| initially be 0.
-    1. While |index| < |list|'s [=list/size=]:
-        1. Invoke |steps| with |list|[|index|].
+    1. While |index| < |list|'s [=list/size=] (including [=list/holes=]):
+        1. Let |item| be |list|[|index|] (including [=list/holes=]).
         1. Increment |index| by 1.
+        1. If |item| is a [=list/hole=], [=iteration/continue=].
+        1. Invoke |steps| with |item|.
 
-    Note: Both the [=list/size=] of |list|, and the order of its items, can potentialy change while
-    invoking |steps|.
+    Note: Entries deleted from the list before being visited will not be processed.
+    Otherwise, every entry in a list is visited exactly once, including any entries added after iteration has begun.
 </div>
 
 <p>To <dfn export for=list,stack,queue,set>clone</dfn> a <a>list</a> |list| is to create a new
 <a>list</a> |clone|, of the same designation, and, <a for=list>for each</a> |item| of |list|,
 <a for=list>append</a> |item| to |clone|, so that |clone| <a for=list>contains</a> the same
-<a for=list>items</a>, in the same order as |list|.
+<a for=list>items</a>, in the same order as |list|, omitting any [=list/holes=] that |list| contains.
 
 <p class=note>This is a "shallow clone", as the <a for=list>items</a> themselves are not cloned in
 any way.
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -1389,6 +1389,7 @@ its <a for=list>size</a> is zero.
 set of steps on each <a for=list>item</a> in order, use phrasing of the form
 "<a for=list>For each</a> |item| of <var ignore>list</var>", and then operate on |item| in the
 subsequent prose.
+The subsequent prose constitutes the |steps| of [=sequence iteration=] over the [=list=].
```
domenic, line 1392: I think this needs to tie to "sequence iteration" more directly. Right now the normative definition of For each is still roughly "performing a set of steps on each [item](https://whatpr.org/infra/451.html#list-item) in order".

Maybe

> To **iterate** over a list _list_ performing steps _steps_ on each item, perform sequence iteration given _list_ and _steps_. This can also be (and usually is) written as "For each _item_ of _list_: _steps_".

Edit: but see below, actually I think trying to share the code between lists and maps doesn't work well.

infra.bs
```diff
@@ -1424,6 +1425,23 @@ ascending order, with |a| being less than |b| if |a|'s second <a for=struct>item
 <a>code unit less than</a> |b|'s second <a for=struct>item</a>, gives the result « (404,
 "<code>Not Found</code>"), (200, "<code>OK</code>"), (null, "<code>OK</code>") ».</p>
 
+<div algorithm>
+    <dfn>Sequence iteration</dfn> over an ordered sequence |seq|,
```
domenic, line 1429: Follow prevailing style and hard-wrap to 100 characters, here and below.

domenic, line 1429: "ordered sequence" is not defined. So I think the desire to share code here between lists and maps doesn't work well. You could assemble a list of map entries and operate on them, but doing so is a lot of work...

I think it might be best to just define it separately, with some duplication.

domenic: Yeah, there's always a fine line between tautological definitions and real definitions in Infra, since we haven't decided to ground everything in first-order logic. For this case I think we're trying to be precise about a specific algorithm so I think we should try to be explicit.

bakkot:
Do note that this is _not_ the same semantics as ES Maps use, because deletion is specified differently. Specifically, for an ES Map, deleting leaves a `~empty~` tombstone rather than actually removing something from the list, so that the indices of all the other items in the list of entires are preserved. Maps as specified here actually remove the underlying item from the list. That means that if, while iterating, you delete something from the map which has already been visited, you will end up skipping some random other item from the map (because all indices past the deleted item will get reduced by 1). That seems like something you might not want.

domenic:
Thanks @bakkot, great catch.

Can you think of any way to specify those semantics without tombstones, which feel a bit awkward? Like... snapshotting the keys, but then doing a separate exists check before actually executing the steps?

bakkot:
Offhand nothing else occurs to me, though you might be able to find something in the literature. If you're trying to get exactly ES Map semantics, using the same specification mechanism as ES is probably going to be easiest.

Snapshotting is going to be tricky. There's a bunch of weird edge cases: like, suppose you had `[a, b, c]`, and the iterator reached and output `b`, and then while it's paused there `a` is removed and re-added - in an ES map the iterator will proceed to output `c` and then `a` again (fun!). That follows immediately from the use of tombstones but I think trying to write that down with snapshots is going to end up being pretty painful.

tabatkins:
Yeah, indeed, I think capturing the precise semantics without using tombstones would be a *huge* burden.

My inclination is to specify that lists and maps can have holes, but they're automatically skipped over by all spec-author-facing operations; only the list/map iteration internal algo sees them.

domenic:
Yeah, that seems fair.

So there are several competing issues here:

- We want spec authors to be able to write Web IDL maplikes/setlikes without doing a bunch of ECMASpeak (https://github.com/whatwg/webidl/issues/824). Two ways of doing this come to mind:
  - Having backing Infra maps/sets, like @tabatkins has done in https://github.com/whatwg/webidl/pull/1138 . Then make sure Infra maps/sets align semantically with JS ones, at least when used in Web IDL contexts. +0/-0 I'm not too worried about, but iteration (discussed here) is the hard part.
  - Create a bunch of wrapper operations around the ECMASpeak, e.g. "maplike set", "setlike add", etc. in Web IDL, which spec authors have to use on their maplikes/setlikes.
  - The former of these options is obviously a bit more user-friendly from a spec author point of view.
- Infra currently does not specify any well-defined iteration behavior (#396). It would be good to specify this regardless of any Web IDL/ES overlap.
  - One potential path here we gestured at in #396 is to just disallow modifications during iteration, i.e., Infra-using specs must not do that or allow author code to do that.
  - That is only viable if we don't have Infra maps/sets back maplikes/setlikes, because maplike/setlike iteration is user-hookable.
  - And it seems pretty unlikely to be viable regardless. It's too easy to call author code while iterating.
  - So, we probably need to pick some real semantics. We have several choices.
- Infra sets are defined as a specialization of Infra lists.
  - However, JS Set and JS Array iteration behavior is not the same; array iteration does not have the tombstone behavior.
  - Do we want to have Infra list and Infra set iteration behavior differ?
  - If we do, then unless there's something really clever we can do in the iteration algorithm, per @bakkot's point above we probably need to rewrite how sets are done to support this tombstone-type semantic.
  - (We'd also need to rewrite maps to support that, but this is less of an issue because maps are hand-wavey "sequences" and not specializations of lists.)
- Infra maps and sets are generally meant to map to C++ map/set classes used in implementations.
  - I don't know what iteration behavior those have in implementations. I kind of doubt it's the same in different browsers.
  - So, there might be existing interop problems with various web platform specs that use Infra maps/sets, if those maps/sets can be modified mid-iteration.

I _think_ the best path forward here is to push toward aligning Infra maps/sets with JS ones, and try to be sure to have people write web platform tests for modifications-during-iteration for any Infra map/set usage (not just maplike/setlike, but other cases too).

But the wrapper operations path is also viable. It would basically defer the question of how to precisely specify Infra map/set iteration until later.

domenic:
(For the record I typed up most of the above, went to a meeting, missed @tabatkins's response, then pressed submit.)

> My inclination is to specify that lists and maps can have holes, but they're automatically skipped over by all spec-author-facing operations; only the list/map iteration internal algo sees them.

Do you mean set/map, or list/map? Changing lists to have holes might have other knock-on effects, e.g. by making list iteration mismatch `Array` iteration.

tabatkins:
I meant list, but I'm not wedded to that; i'd be okay with lists and sets having different iteration behavior if needed.

Regarding Array iteration, note that Arrays already differ from themselves between forEach() and iteration, so who knows what we should be consistent with.

domenic:
Fair point. It's probably fine to add holes to lists; I'm happy for you to give it a shot.

annevk:
I wonder to what extent we should give advice here to implementations as not all lists will need this complexity. I wonder if we can do similar things to "ASCII string" whereby we essentially assert the thing will have a subset of the total semantics.

tabatkins:
Okay, I won't get back to finishing this PR until at least next week due to vacation, but just gathering some thoughts here:

1. Holes are possible. I don't want normal uses to require thinking about holes, so either I specify that they're automatically skipped by all operations *other than* iteration.

2. Holes aren't possible for normal lists/etc, but I go with Anne's suggestion and define a specialization of lists/etc that can have holes, and use *those* for maplike/setlike. These would generally be necessary for *any* spec that wants to have mutation during an iteration; mutation during iteration will be explicitly forbidden for normal lists/etc. Holes will still only be visible for Infra-internal purposes; spec authors will never see them.

3. Holes aren't possible; instead, I describe iteration in terms of linked lists. I am *pretty sure* that this gives the same observable behavior as using holes, without requiring me to acknowledge the existence of holes. However, a literal implementation probably isn't very efficient, and carefully exposing the right behavior when using a different impl strategy would be non-trivial.

I think I don't like (3). I'm leaning to (2), but could be convinced on (1) if we think it's better to put the burden of thinking about which type to use on impls rather than spec authors. We'd have guidance either way in Infra, just targeted at different audiences.

domenic:
@tabatkins ping on picking this back up!

I think I prefer (1) if possible, but I'm OK with (2) if you prefer.

rakuco:
Hi from Blink-land. I've been trying to review @yuki3's patches making Blink's pair iterators and map/setlike code more spec-conformant and ended up stumbling upon #396 and this PR. It would be great to have more clarity on this; I was wondering if there's been any progress with the patch and the approaches discussed above.

tabatkins:
It's remained on my list of things to wrap up "real soon now". :/

tabatkins:
Okay, picking this back up now. I've gone with Option 1 from [my comment](https://github.com/whatwg/infra/pull/451#issuecomment-1122822681) - lists and maps can have holes, which are invisible to all algorithms unless explicitly specified. This is done in a somewhat handwavey style that is *hopefully* okay, but if we want to be more precise I can add an explicit "with holes" argument to every algo, defaulting to false, and pass that around appropriately.

Big thing left is to just decide which of the list iteration behaviors to steal from JS: Array.forEach records the length of the list at the start and stops at that length, even if items were added mid-iteration; Array iteration recalculates the length on each iteration to check if it should stop, so items added mid-iteration will be visited as well.  JS Maps do the latter behavior in both instances, so I'm currently leaning towards (and the spec text reflects) doing that for lists as well.

zcorpan:
This came up in https://github.com/WICG/sanitizer-api/pull/375

This PR looks OK to me, but I'm not too familiar with the details here.

PR #452, opened 2022-05-05
Add example of code-unit-less-than being counterintuitive. Fixes #388

### LINKED ISSUE #388: Add counterintuitive examples of code unit less than
@wanderview and I spent a morning trying to figure out whether [code unit less than](https://infra.spec.whatwg.org/#code-unit-less-than) matches code point comparison (or, equivalently, UTF-8 byte comparison). It does not.

The reason is surrogates. But at first I thought it was only unpaired surrogates, so that if you had two USVStrings you would be safe. @dbaron pointed out that was incorrect and you can get different results for code unit vs. code point sorting even with USVStrings.

We should update Infra with at least the following examples, plus probably a few less exciting ones. https://icu-project.org/docs/papers/utf16_code_point_order.html has some good info and examples as well.

- U+FF61 is "code unit greater than" U+10002 since 0xFF61 > 0xD800 0xDC02, even though it is "code point less than".
- Similarly U+FB00 (ﬀ) is "code unit greater than" U+10478 (𐑸) since 0xFB00 > 0xD801 0xDC78, even though it is "code point less than".

tabatkins:
Added an example to the note, utilizing fullwidth tilde (a single code unit) and smiley (higher codepoint, but two code units).

***
<a href="https://whatpr.org/infra/452.html" title="Last updated on May 6, 2022, 7:53 PM UTC (616d16f)">Preview</a> | <a href="https://whatpr.org/infra/452/3deb8ac...616d16f.html" title="Last updated on May 6, 2022, 7:53 PM UTC (616d16f)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -987,9 +987,10 @@ string, producing a highly efficient, consistent, and deterministic sort order. 
 ordering will not match any particular alphabet or lexicographic order, particularly for
 <a>code points</a> represented by a surrogate pair. [[!ECMA-262]]
 
-<p class="example" id="example-code-unit-less-than">For example, the code point U+FF5E FULLWIDTH TILDE (～) is obviously less than the
-code point U+1F600 (😀), but the tilde is composed of a single code unit 0xFF5E, while the smiley is
-composed of two code units 0xD83D and 0XDE00, so the smiley is [=code unit less than=] the tilde.
+<p class="example" id="example-code-unit-less-than">For example, the code point U+FF5E FULLWIDTH
+TILDE (～) is obviously less than the code point U+1F600 (😀), but the tilde is composed of a single
+code unit 0xFF5E, while the smiley is composed of two code units 0xD83D and 0XDE00, so the smiley is
+[=code unit less than=] the tilde.
 
 <hr>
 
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -985,7 +985,10 @@ literal, we can use plainer language: <var>domain</var> <a for=string>ends with<
 {{Array/sort()}} method on an array of strings. This ordering compares the 16-bit code units in each
 string, producing a highly efficient, consistent, and deterministic sort order. The resulting
 ordering will not match any particular alphabet or lexicographic order, particularly for
-<a>code points</a> represented by a surrogate pair. [[!ECMA-262]]
+<a>code points</a> represented by a surrogate pair. For example, the codepoint U+FF5E
+(～ FULLWIDTH TILDE) is obviously less than the codepoint U+1F600 (😀), but the tilde is
```
domenic, line 989: "codepoint" -> "code point" in two places

Per https://infra.spec.whatwg.org/#code-points the exact notation should be `U+FF5E FULLWIDTH TILDE (～)`

infra.bs
```diff
@@ -985,7 +985,10 @@ literal, we can use plainer language: <var>domain</var> <a for=string>ends with<
 {{Array/sort()}} method on an array of strings. This ordering compares the 16-bit code units in each
 string, producing a highly efficient, consistent, and deterministic sort order. The resulting
 ordering will not match any particular alphabet or lexicographic order, particularly for
-<a>code points</a> represented by a surrogate pair. [[!ECMA-262]]
+<a>code points</a> represented by a surrogate pair. For example, the codepoint U+FF5E
```
domenic, line 988: Optional: you could split this from the note `<p>` and put it in a separate `<p class="example">`

domenic (commented):
Looks good with nits, thanks very much. This helps a lot.

domenic (approved):

PR #491, opened 2022-10-21
Recommend moments, durations, and advice from hr-time.

### LINKED ISSUE #280: Consider adding time and duration types
https://wicg.github.io/BackgroundSync/spec/PeriodicBackgroundSync-index.html has a couple terms, like [minimum periodic sync interval for any origin](https://wicg.github.io/BackgroundSync/spec/PeriodicBackgroundSync-index.html#minimum-periodic-sync-interval-for-any-origin), that it currently defines as "a long long [number of milliseconds]". However, this number isn't exposed to Javascript anywhere, so it doesn't really need a particular data type other than being an interval of time. It might be easier to get specs like this right if Infra defined [types](https://infra.spec.whatwg.org/#primitive-data-types) for points in time and amounts of time.

jyasskin:
Fixes #280.

I picked "instant" and "duration" to match the [Temporal](https://tc39.es/proposal-temporal/docs/instant.html) [types](https://tc39.es/proposal-temporal/docs/duration.html), but I'd be equally happy with "point in time" and "length of time" or any other choice y'all prefer.

***
<a href="https://whatpr.org/infra/491.html#time" title="Last updated on Feb 23, 2023, 9:59 AM UTC (b756548)">Preview</a> | <a href="https://whatpr.org/infra/491/eb68213...b756548.html" title="Last updated on Feb 23, 2023, 9:59 AM UTC (b756548)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -1311,10 +1311,12 @@ these steps:
 <p class=example id=example-string-concatenate>To serialize a set <var>set</var>, return the
 <a for=string>concatenation</a> of <var>set</var> using U+0020 SPACE.
 
+
 <h3 id=time>Time</h3>
 
-Represent time using the [=moment=] and [=duration=] types. Follow the advice in
-[[HR-TIME-3#sec-tools]] when creating these and exchanging them with JavaScript.
+<p>Represent time using the [=moment=] and [=duration=] specification types. Follow the advice in
+[[HR-TIME-3#sec-tools]] when creating these and exchanging them with JavaScript. [[HR-TIME]]
+
 
 <h2 id=data-structures>Data structures</h2>
 
--- a/infra.bs
+++ b/infra.bs
@@ -1311,6 +1311,10 @@ these steps:
 <p class=example id=example-string-concatenate>To serialize a set <var>set</var>, return the
 <a for=string>concatenation</a> of <var>set</var> using U+0020 SPACE.
 
+<h3 id=time>Time</h3>
+
+Represent time using the [=moment=] and [=duration=] types. Follow the advice in
+[[HR-TIME-3#sec-tools]] when creating these and exchanging them with JavaScript.
 
 <h2 id=data-structures>Data structures</h2>
 


infra.bs
```diff
@@ -1296,6 +1297,29 @@ these steps:
...
+<p>A <dfn export>duration</dfn> is a length of time, the distance from one
+[=instant=] to another one (or the sum of several such distances). [=Durations=]
+are signed: the [=duration=] from a later [=instant=] to an earlier one is
+negative.
+
+<div class=example id="example-times">
+ <p><code>new {{Date}}(<var>value</var>)</code> represents the [=instant=] that
+ is <var>value</var> milliseconds after the [=Unix epoch=].
+
+ <p><code>performance.{{Performance/now}}()</code> returns the [=duration=]
```
domenic, line 1318:
```suggestion
 <p>{{Performance/now()|performance.now()}} returns the [=duration=]
```

infra.bs
```diff
@@ -1296,6 +1297,29 @@ these steps:
...
+algorithm executes the step using [=now=]. Two different uses of [=now=] return
+different [=instants=], so most algorithms should save it into a variable once,
+and use that variable multiple times.
+
+<p>A <dfn export>duration</dfn> is a length of time, the distance from one
+[=instant=] to another one (or the sum of several such distances). [=Durations=]
+are signed: the [=duration=] from a later [=instant=] to an earlier one is
+negative.
+
+<div class=example id="example-times">
```
domenic, line 1314:
```suggestion
<div class=example id=example-times>
```

domenic (commented):
I like this in general. We should think through the integration story with other specs, particularly performance timing specs.

Can you tell us more about where you want to use it?

We should probably add some note explaining that if the time is exposed to developers, you likely want to use the stuff in High Resolution Time (e.g. relative high resolution time).

How does this relate to https://w3c.github.io/hr-time/#sec-monotonic-clock , and to monotonic time in general? In particular, can multiple invocations of [=now=] ever end up in a situation where the second one returns a value below the first?

I guess "[DOMHighResTimeStamp](https://w3c.github.io/hr-time/#dom-domhighrestimestamp) representing the high resolution time at which the [shared monotonic clock](https://w3c.github.io/hr-time/#dfn-shared-monotonic-clock) is zero" is supposed to be the Unix epoch?

And I guess "[unsafe shared current time](https://w3c.github.io/hr-time/#dfn-unsafe-shared-current-time)" might be [=now=]?

annevk:
This seems reasonable as an underpinning for High Resolution Time and specifications that need time without exposing it to script. We should point out that for script-exposed times https://w3c.github.io/hr-time/#dfn-current-high-resolution-time and friends are to be used instead.

cc @noamr @yoavweiss

noamr:
> I guess "[DOMHighResTimeStamp](https://w3c.github.io/hr-time/#dom-domhighrestimestamp) representing the high resolution time at which the [shared monotonic clock](https://w3c.github.io/hr-time/#dfn-shared-monotonic-clock) is zero" is supposed to be the Unix epoch?

No, zero-monotonic is implementation/OS specific, and has an offset from unix epoch which (in chrome) is constant for the duration of the browser runtime.

jyasskin:
I did add the advice to use

annevk (commented):
I'm not entirely sure what the best thing to do here is. It seems those two example specifications would end up exposing the time at which point resolution becomes somewhat important. Maybe we should stress that instants and durations when serialized in some manner cannot be more accurate than seconds?

infra.bs
```diff
@@ -1296,6 +1297,40 @@ these steps:
...
+
+<p>A <dfn export>duration</dfn> is a length of time, the distance from one
+[=instant=] to another one (or the sum of several such distances). [=Durations=]
+are signed: the [=duration=] from a later [=instant=] to an earlier one is
+negative.
+
+When exchanging [=instants=] and [=durations=] with Javascript, use
+{{DOMHighResTimeStamp}}, which represents a [=duration=] as a {{double}} number
+of milliseconds. To represent an [=instant=] as a {{DOMHighResTimeStamp}}, use
+the duration from the [=Unix epoch=] to that [=instant=].
```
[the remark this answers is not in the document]
yoavweiss: +1 to what @annevk said.

DOMHighResTimestamp is relative to the [time origin](https://www.w3.org/TR/hr-time-3/#dfn-time-origin). 
What would this infrastructure be used for?

infra.bs
```diff
@@ -1296,6 +1297,40 @@ these steps:
...
+ <p><code>new {{Date}}(<var>value</var>)</code> represents the [=instant=] that
+ is <var>value</var> milliseconds after the [=Unix epoch=].
+
+ <p>{{Performance/now()|performance.now()}} returns the [=duration=]
+ from the [=environment settings object/time origin=] to [=now=],
+ represented as a {{double}} number of milliseconds, with its resolution limited
+ by the [=coarsen time=] algorithm.
+</div>
+
+Time gets more complicated the longer you study it. It's possible for a use of
```
yoavweiss, line 1329: Is this a note?

yoavweiss (commented):
A couple of comments. My initial reaction is that it might be better to augment HR-time with whatever definitions are required (and potentially rename it to reflect that, as it's already defining more than just DOMHighResTimestamp)

jyasskin:
I'm working on putting this in HR Time instead, and making sure the concepts are rigorous to capture everything there. I think we'll want a pointer from Infra to those definitions so that people find them, but I'll let you know when that's ready.

jyasskin:
I haven't yet updated this PR to point to my proposal in HR-Time, but please check out https://pr-preview.s3.amazonaws.com/jyasskin/hr-time/pull/140.html#sec-tools and let me know what you think in w3c/hr-time#140.

jyasskin:
w3c/hr-time#140 is merged, so please take a look at this new reference to it.

annevk (approved):
Apologies for the delay here. I think this is okay. I'm a little weary as we probably wouldn't mention URLs here, for instance, but arguably time is fundamental enough to warrant a mention without an actual local definition.

I'd like @domenic to have another look as well before this is merged.

Note that I pushed some nits as well.

jyasskin:
Thanks for the nit fixes!

domenic (approved):
Seems good; thanks for working on this Jeffrey!!

PR #495, opened 2022-10-26
Meta: my employer changed

annevk:


domenic (approved):
Can we just do "Apple", like we did "Mozilla" and still do "Google"? No worries if that's not possible, but it seems a bit cleaner.

Issue #508, opened 2022-12-13
Document multiple assignment tuple unwrapping


dlrobertson:
"Let (_variableOne_, _variableTwo_) be the result of _functionReturningTuple_" is a valid statement, but is not documented in [Tuples](https://infra.spec.whatwg.org/#tuples).

_Originally posted by @domenic in https://github.com/whatwg/dom/issues/1134#issuecomment-1347741093_
> We've been doing it as "Let (_prefix_, _localName_) be the result of..." but only for tuples, not lists. (And it's not documented anywhere.)

dlrobertson:
Done! Opened https://github.com/whatwg/infra/issues/509 and added context

PR #512, opened 2022-12-18
Add documentation for tuple multiple assignment

dlrobertson:
"Let (variableOne, variableTwo) be the result of functionReturningTuple" is a valid statement, document multiple asignment usage and write a basic example.

Fixes: #508

***
<a href="https://whatpr.org/infra/512.html" title="Last updated on Dec 21, 2022, 6:28 AM UTC (c8e206d)">Preview</a> | <a href="https://whatpr.org/infra/512/171a136...c8e206d.html" title="Last updated on Dec 21, 2022, 6:28 AM UTC (c8e206d)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -427,6 +427,21 @@ behavior; otherwise null.
 <a href=https://en.wikipedia.org/wiki/Scope_(computer_science)#Block_scope>block scoped</a>.
 Variables must not be declared more than once per algorithm.
 
+<p>A multiple assignment syntax can be used to assign multiple variables to the <a>tuple</a>'s
+<a for=tuple>items</a>, by surrounding the variable names with parenthesis and separating each
+variable name by a comma. The number of variables assigned cannot differ from the number of
+<a for=tuple>items</a> in the <a>tuple</a>.
+
+<div class=example id=example-tuple-multiple-assignment>
+ <ol>
+  <li><p>Let |statusInstance| be the status (200, `<code>OK</code>`).
+  <li><p>Let (|status|, |statusMessage|) be |statusInstance|.
+ </ol>
+
+ <p>Assigning |status| and |statusMessage| could be written as two separate steps that use an index
+ or <a for=tuple>name</a> to access the <a>tuple</a>'s <a for=tuple>items</a>.
+</div>
+
 
 <h3 id=algorithm-control-flow>Control flow</h3>
 


infra.bs
```diff
@@ -1669,6 +1669,21 @@ to the <a>tuple</a>. An indexing syntax can be used by providing a zero-based in
 Infra Standard might need the flexibility to add new <a for=struct>names</a> to their struct
 without breaking literal syntax used by their dependencies. In that case a tuple is not appropriate.
 
+A multiple assignment syntax can be used to assign multiple variables to the <a>tuples</a>
```
annevk, line 1672:
```suggestion
A multiple assignment syntax can be used to assign multiple variables to the <a>tuple</a>'s
```

infra.bs
```diff
@@ -1669,6 +1669,21 @@ to the <a>tuple</a>. An indexing syntax can be used by providing a zero-based in
...
+<a for=tuple>items</a> in the <a>tuple</a>.
+
+<div class=example id=example-tuple-multiple-assignment>
+ <ol>
+  <li><p>Let |statusInstance| be the status (200, `<code>OK</code>`).
+  <li><p>Let (|status|, |statusMessage|) be |statusInstance|.
+ </ol>
+
+ <p>Assigning |status| and |statusMessage| could be written as two separate steps that use an index
+ or <a for=tuple>name</a> to access the <a>tuples</a> <a for=tuple>item</a>.
```
annevk, line 1684:
```suggestion
 or <a for=tuple>name</a> to access the <a>tuple</a>'s <a for=tuple>items</a>.
```

infra.bs
```diff
@@ -1648,7 +1648,7 @@ need to be clear from context. This can be done by preceding the first instance
 to the <a>tuple</a>. An indexing syntax can be used by providing a zero-based index into a
 <a>tuple</a> inside square brackets. The index cannot be out-of-bounds.
 
-<div class=example id=example-tuple>
+<div class=example id=example-tuple-construction>
```
annevk, line 1651: Let's not change the ID here. Even if it would be slightly better looking at it now, it's generally preferable to keep IDs working so links people might have exchanged don't stop working.
```suggestion
<div class=example id=example-tuple>
```

annevk (commented):
I think we should probably move this to the Variables section, but otherwise this looks good to me. @domenic you have any thoughts?

infra.bs
dlrobertson: :+1: Yeah that makes sense

annevk (approved):

PR #520, opened 2023-01-09
Address limits on unconstrained inputs

annevk:
The second half of this paragraph is from the HTML Standard.

***
<a href="https://whatpr.org/infra/520.html" title="Last updated on Jan 16, 2023, 8:20 AM UTC (c645968)">Preview</a> | <a href="https://whatpr.org/infra/520/eb68213...c645968.html" title="Last updated on Jan 16, 2023, 8:20 AM UTC (c645968)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -239,12 +239,38 @@ paths would be an insurmountable task and not productive as they would not stand
 as well as the single algorithm would. Therefore performance is best left as a field to compete
 over.
 
-<p tracking-vector>A document using the Infra Standard should not enforce specific limits on
-unconstrained inputs to allow for competition among user agents and foster ever greater data
-structures over time. I.e., avoid constraining the computing needs of the future. Nevertheless, user
-agents may impose <a>implementation-defined</a> limits on otherwise unconstrained inputs. E.g., to
-prevent denial of service attacks, to guard against running out of memory, or to work around
-platform-specific limitations.
+<h4 id=algorithm-conformance-limits>Avoid limits on algorithm inputs</h4>
+
+<p>A document using the Infra Standard generally should not enforce specific limits
+on algorithm inputs with regards to their size, resource usage, or equivalent. This allows for
+competition among user agents and avoids constraining the potential computing needs of the future.
+
+<p tracking-vector>Nevertheless, user agents may impose <a>implementation-defined</a> limits on
+otherwise unconstrained inputs. E.g., to prevent denial of service attacks, to guard against running
+out of memory, or to work around platform-specific limitations.
+
+<p class=note>Global resource limits can used as side channels through a variant on a resource
+exhaustion attack whereby the attacker can observe whether a victim application reaches the global
+limit. Limits could also be used to fingerprint the user agent, but only if they make the user agent
+more unique in some manner, e.g., if they are specific to the underlying hardware.
+
+<p class=example id=example-algorithm-conformance-limits>An API that allows creating an in-memory
+bitmap might be specified to allow any dimensions, or any dimensions up to some large limit like
+JavaScript's <code>Number.MAX_SAFE_INTEGER</code>. However, implementations can choose to impose
+some <a>implementation-defined</a> (and thus not specified) limit on the dimensions, instead of
+attempting to allocate huge amounts of memory.
+
+<p class=example id=example-algorithm-conformance-limits-language>A programming language might not
+have any maximum call stack size specified. However, implementations could choose to impose one for
+practical reasons.
+
+<p>There are at least two cases where a limit can be useful to define:
+
+<ul class=brief>
+ <li>Ensuring all implementations can handle inputs of a given minimum size, i.e., a lower limit.
+ <li>Interoperability: code can end up depending on particular limits and sometimes embracing that
+ is not problematic for the future and can make the code run in more user agents.
+</ul>
 
 
 <h3 id=algorithm-declaration>Declaration</h3>
--- a/infra.bs
+++ b/infra.bs
@@ -268,8 +268,8 @@ practical reasons.
 
 <ul class=brief>
  <li>Ensuring all implementations can handle inputs of a given minimum size, i.e., a lower limit.
- <li>Interoperability: code can end up depending on particular limits and sometimes embracing that
- is not problematic for the future and can make the code run in more user agents.
+ <li>Interoperability: code can end up depending on particular limits. Sometimes, embracing that
+ is not problematic for the future, and can make the code run in more user agents.
 </ul>
 
 
--- a/infra.bs
+++ b/infra.bs
@@ -250,8 +250,8 @@ user agents and avoids constraining the potential computing needs of the future.
 otherwise unconstrained inputs. E.g., to prevent denial of service attacks, to guard against running
 out of memory, or to work around platform-specific limitations.
 
-<p class=note>Global resource limits can used as side channels through a variant on a resource
-exhaustion attack whereby the attacker can observe whether a victim application reaches the global
+<p class=note>Global resource limits can be used as side channels through a variant on a resource
+exhaustion attack, whereby the attacker can observe whether a victim application reaches the global
 limit. Limits could also be used to fingerprint the user agent, but only if they make the user agent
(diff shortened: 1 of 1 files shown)


annevk:
@jyasskin care to review as well?

jasnell (approved):

infra.bs
```diff
@@ -239,6 +239,13 @@ paths would be an insurmountable task and not productive as they would not stand
 as well as the single algorithm would. Therefore performance is best left as a field to compete
 over.
 
+<p tracking-vector>A document using the Infra Standard should not enforce specific limits on
+unconstrained inputs to allow for competition among user agents and foster ever greater data
```
jyasskin, line 243: The flip side of competition among user agents is that web pages might exceed one browser's limits while only testing on a larger browser. We should probably check with a Firefox person that they're ok with that problem in this case?

jyasskin:
Oh, and even if we discourage exact limits, I think we should allow minimum sizes on the implementation-defined limits, like https://httpwg.org/specs/rfc9110.html#uri.references recommends for URIs.

infra.bs
```diff
@@ -239,6 +239,13 @@ paths would be an insurmountable task and not productive as they would not stand
 as well as the single algorithm would. Therefore performance is best left as a field to compete
 over.
 
+<p tracking-vector>A document using the Infra Standard should not enforce specific limits on
```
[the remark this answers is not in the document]
annevk: We don't really have a good solution yet for global limits, do we? But I agree that a note about that would be useful.

annevk:
@smaug---- @martinthomson care to comment on behalf of Firefox as suggested above?

infra.bs
annevk: Or pages rely on a particular limit. The note should probably talk about this sometimes leads to a particular limit being defined.

Another thing I realized is that by not addressing limits we don't define the error handling for hitting them either. At some point there was some ambition to address that for out-of-memory in JS, but not enough to actually make it happen.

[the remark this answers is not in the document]
smaug----: I like what Domenic suggests. And with an example 'allows for competition among user agents' part feels less strange.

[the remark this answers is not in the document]
martinthomson: SGTM too.  Implementations need to determine how they allocate the resources necessary and so are free to set limits.

What would be really good is a little more about how specific limits are either dictated by implementation choices (which don't generally vary between installations of the same implementation and so don't contribute much to tracking), by hardware (which can vary and are consequently a source of tracking information), or both (with variable effect on tracking).  Implementations are might in a position to limit how much information is revealed through limits, such as through quantization.  This might involve sacrificing some potential performance as the applied limits might be smaller than what hardware is capable of, which puts the privacy goal in tension with any desire to maximize performance.  Again, this is an area of implementation differentiation as different approaches seek different trade-offs.

(In protocol land, there are often resource allocations that are made by others or jointly made, in which case this loose sort of approach doesn't work, but that applies on a case-by-case basis and probably isn't worth expounding upon.)

infra.bs
```diff
@@ -239,6 +239,39 @@ paths would be an insurmountable task and not productive as they would not stand
 as well as the single algorithm would. Therefore performance is best left as a field to compete
 over.
 
+<h4 id=algorithm-conformance-limits>Avoid limits on algorithm inputs</h4>
```
domenic, line 242: I think this should be a h3 since it's not really about the must/may/etc. stuff that the "Conformance" section is mostly about.

infra.bs
```diff
@@ -239,6 +239,39 @@ paths would be an insurmountable task and not productive as they would not stand
...
+<p class=example id=example-algorithm-conformance-limits-language>A programming language might not
+have any maximum call stack size specified. However, implementations could choose to impose one for
+practical reasons.
+
+<p>There are at least two cases where a limit can be useful to define:
+
+<ul class=brief>
+ <li>Ensuring all implementations can handle inputs of a given minimum size, i.e., a lower limit.
+ <li>Interoperability: code can end up depending on particular limits and sometimes embracing that
+ is not problematic for the future and can make the code run in more user agents.
```
domenic, line 272:
```suggestion
 <li>Interoperability: code can end up depending on particular limits. Sometimes, embracing that
 is not problematic for the future, and can make the code run in more user agents.
```

domenic (approved):
LGTM with nits

annevk:
Anyone want to do another pass? I'll merge this Monday if there's no further feedback.

infra.bs
```diff
@@ -240,6 +240,40 @@ as well as the single algorithm would. Therefore performance is best left as a f
...
+
+<p>A document using the Infra Standard generally should not enforce specific limits on algorithm
+inputs with regards to their size, resource usage, or equivalent. This allows for competition among
+user agents and avoids constraining the potential computing needs of the future.
+
+<p tracking-vector>Nevertheless, user agents may impose <a>implementation-defined</a> limits on
+otherwise unconstrained inputs. E.g., to prevent denial of service attacks, to guard against running
+out of memory, or to work around platform-specific limitations.
+
+<p class=note>Global resource limits can used as side channels through a variant on a resource
```
das7pad, line 253:
```suggestion
<p class=note>Global resource limits can be used as side channels through a variant on a resource
```
Nit: It looks like there is a missing 'be'; and perhaps also a missing comma before 'whereby' below?

infra.bs
```diff
@@ -240,6 +240,40 @@ as well as the single algorithm would. Therefore performance is best left as a f
...
 
+<h3 id=algorithm-limits>Avoid limits on algorithm inputs</h3>
+
+<p>A document using the Infra Standard generally should not enforce specific limits on algorithm
+inputs with regards to their size, resource usage, or equivalent. This allows for competition among
+user agents and avoids constraining the potential computing needs of the future.
+
+<p tracking-vector>Nevertheless, user agents may impose <a>implementation-defined</a> limits on
+otherwise unconstrained inputs. E.g., to prevent denial of service attacks, to guard against running
+out of memory, or to work around platform-specific limitations.
```
jyasskin, line 251: Inspired by https://github.com/whatwg/infra/pull/520#discussion_r1065439589:
```suggestion
out of memory, or to work around platform-specific limitations. Specifications should define what
happens when input exceeds such an <a>implementation-defined</a> limit.
```

I'm not sure that's exactly the right wording or location.

infra.bs
```diff
@@ -240,6 +240,40 @@ as well as the single algorithm would. Therefore performance is best left as a f
...
+attempting to allocate huge amounts of memory.
+
+<p class=example id=example-algorithm-limits-language>A programming language might not have a
+maximum call stack size specified. However, implementations could choose to impose one for practical
+reasons.
+
+<p>There are at least two cases where a limit can be useful to define:
+
+<ul class=brief>
+ <li>Ensuring all implementations can handle inputs of a given minimum size, i.e., a lower limit.
```
jyasskin, line 271: This is a bit awkward. Maybe:
```suggestion
 <li>Constraining an <a>implementation-defined</a> limit to be at least a particular minimum size.
```
Even this wording isn't really a case "where a limit can be useful to define", so maybe it should move out of this list.

jyasskin (approved):
2 minor suggestions:

infra.bs
annevk: I don't think we can require this. At the very least it should be a follow-up discussion. E.g., consider out-of-memory. Currently implementations can impose limits and throw an exception or they can let the entire process crash. We wouldn't want specifications to constrain that.

infra.bs
annevk: Why not? Isn't a lower limit a limit?

infra.bs
jyasskin: It's already "should" rather than "must", but maybe with "if doing so is reasonable"? I agree that we won't be able to do it in all cases, but I wanted to encourage people to try. Even in the case of memory limits, something should specify the type of error thrown if one is thrown.

I have no problem making that a follow up discussion.

annevk: #523.

infra.bs
annevk: Fair enough, reworded.

Issue #518, opened 2023-01-06
ECMA262 completion records are not interpreted appropriately


jugglinmike:
ECMA262 uses a control flow convention built on [completion records](https://tc39.es/ecma262/#sec-completion-record-specification-type)--"wrapper" values that the standard's macro-like shorthands (e.g. [ReturnIfAbrupt](https://tc39.es/ecma262/#sec-returnifabrupt) and [the question mark symbol](https://tc39.es/ecma262/#sec-returnifabrupt-shorthands)) use to conditionally halt algorithms.

The effect is similar to the notion of "exception throwing" in web specifications, but the mechanism is fundamentally different. This means that fallible ECMA262 algorithms cannot be interpreted identically in web specs.

For instance, consider ["serialize a JavaScript value to a JSON string" in Infra](https://infra.spec.whatwg.org/#serialize-a-javascript-value-to-a-json-string)

> 1. Let result be ? Call(%JSON.stringify%, undefined, « value »).
> 1. If result is undefined, then throw a TypeError.
> 1. Assert: result is a string.
> 1. Return result.

According to the definition of the `?` shorthand, the first step either stores a value in "result" or causes this algorithm to return a "throw" completion record. Combining those possibilities with the following steps means that the algorithm can have one of three results: returning a "throw" completion record, throwing an error, or returning a string.

In ECMA262, [the first two are different ways of saying the same thing](https://tc39.es/ecma262/#sec-throw-an-exception), but that isn't the case in web specs. We can see this in the call site of the algorithm described above:

> 1. Let string be the result of [serializing a JavaScript value to a JSON string](https://infra.spec.whatwg.org/#serialize-a-javascript-value-to-a-json-string) given value.

Were this written in ECMA262, that invocation would likely be preceded by the `?` symbol. The Infra standard probably omits it because web specifications rely on exception throwing semantics. However, the effect is that the value of the variable named "string" may be a "throw" completion record.

Although there are only three instances of the `?` operator in Infra, [I'm told that this pattern also shows up in HTML and in WebIDL](https://matrix.to/#/!AGetWbsMpFPdSgUrbs:matrix.org/$8qiOkJIDoKchnNXTOKiMjKjwrdjJFU26D7w-fFhz3Us?via=matrix.org&via=mozilla.org&via=igalia.com), so I'm reporting this issue to start a discussion about the best way to address the problem.

jugglinmike:
[...]

PR #539, opened 2023-02-16
Introduce a shorthand for ECMA262 abstract ops

jugglinmike:
This patch is in service of gh-518.

I've implemented this using the definition name "throw-if-abrupt" because I can't get Bikeshed to use `?` (U+003F) as a name. If we think this might limit adoption, I can inquire about modifying Bikeshed.

The new text (which I've largely copied from ECMA262) reduces ambiguity, but it does not eliminate it. There are ways of applying `?` which do not match the example, even within Infra itself:

>     1. Return ? Call(%JSON.parse%, undefined, « string »).

...is *not* "otherwise equivalent to":

>     1. Let |result| be ? AbstractOperation().

For what it's worth, ECMA262 suffers from the same deficiency, though [some folks have been trying to address it](https://github.com/tc39/ecma262/pull/1573). @annevk do you feel that the ambiguity undermines the goal of this patch?

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -14,8 +14,10 @@ urlPrefix: https://tc39.github.io/ecma262/#; spec: ECMA-262;
     type: dfn
         text: %JSON.parse%; url: sec-json.parse
         text: %JSON.stringify%; url: sec-json.stringify
+        text: Completion Record; url: sec-completion-record-specification-type
         text: List; url: sec-list-and-record-specification-type
         text: The String Type; url: sec-ecmascript-language-types-string-type
+        text: abrupt completion; url: sec-completion-record-specification-type
         text: realm; url: realm
     type: method; for: Array; text: sort(); url: sec-array.prototype.sort
     type: abstract-op;
@@ -641,6 +643,35 @@ Standard that should be reported and addressed.
  <ol>
   <li><p>Let |x| be "<code>Aperture Science</code>".
   <li><p><a>Assert</a>: |x| is "<code>Aperture Science</code>".
+</div>
+
+
+<h3 id=ecma-262-abstract-operations>ECMA-262 Abstract Operations</h3>
+
+<p>ECMA-262 uses a specification type named [=Completion Record=] to model exception-based control
+flow. Because <a href=#algorithm-control-flow>web specifications have first-class support for exception handling</a>,
+special care must be taken when interpreting the results of ECMA-262 abstract operations.
+[[!ECMA-262]]
+
+<p>This specification defines a shorthand named <dfn export lt=throw-if-abrupt>?</dfn> to interpret
+ECMA-262 completion records in terms of web specifications' "throw" semantics. Algorithm steps that
+say or are otherwise equivalent to:
+
+<div class=example id=example-throw-if-abrupt>
+  <ol>
+    <li><p>Let |result| be <a lt=throw-if-abrupt>?</a> AbstractOperation().
+  </ol>
+</div>
+
+<p>mean the same thing as:
+
+<div class=example id=example-throw-if-abrupt-expanded>
+  <ol>
+    <li><p>Let |hygienicTemp| be AbstractOperation().
+    <li><p><a>Assert</a>: |hygienicTemp| is a [=Completion Record=].
+    <li><p>If |hygienicTemp| is an [=abrupt completion=], throw |hygienicTemp|.\[[Value]].
+    <li><p>Let |result| be |hygienicTemp|.\[[Value]].
+  </ol>
 </div>
 
 


annevk:
Would be interested in feedback from @jmdyck on this.

infra.bs
```diff
@@ -644,6 +646,35 @@ Standard that should be reported and addressed.
 </div>
 
 
+<h3 id=ecma-262-abstract-operations>ECMA-262 Abstract Operations</h3>
+
```
jmdyck, line 650: I think you need to ease into the topic...
- ECMA-262 has these things called "abstract operations".
- Algorithms in Infra-based specs can invoke them.
- The style of invoking them is different from how you invoke Infra-based algs.
- Something about the correspondence between Infra values and 262 values.
- Sometimes, the return value is a 'completion record'...

infra.bs
```diff
@@ -644,6 +646,35 @@ Standard that should be reported and addressed.
...
+<p>ECMA-262 uses a specification type named [=Completion Record=] to model exception-based control
```
jmdyck, line 651: Completion records aren't *just* for modelling exceptions-based flow; even if ECMAScript didn't have exceptions, it would still need completion records (or something like them) to model break+continue+return control flow.

jmdyck:
As for the problems with defining `?`, I probably don't have anything to say here beyond what I've said in https://github.com/tc39/ecma262/pull/1573

jugglinmike:
Thank you, @jmdyck! I appreciate your encouragement to explain this more holistically. I couldn't find a way to expand the reasoning exactly according to the points you've outlined, but I hope the prose I've pushed up is sufficiently coherent, anyway.

> Completion records aren't _just_ for modelling exceptions-based flow; even if ECMAScript didn't have exceptions, it would still need completion records (or something like them) to model break+continue+return control flow.

Fair point. Those details don't seem necessary to explain the motivation or usage of this shorthand, though, so I've true to accommodate them simply by writing in less sweeping terms ("to model control flow interruptions such as thrown exceptions.")

> As for the problems with defining `?`, I probably don't have anything to say here beyond what I've said in [tc39/ecma262#1573](https://github.com/tc39/ecma262/pull/1573)

After some more thought, I'm feeling like replicating 262's current (though imprecise) language still represents a meaningful step forward here. That said, I've also added a note to surface this consideration and help folks recognize when an improvement is available.

infra.bs
```diff
@@ -648,14 +649,20 @@ Standard that should be reported and addressed.
 
 <h3 id=ecma-262-abstract-operations>ECMA-262 Abstract Operations</h3>
 
-<p>ECMA-262 uses a specification type named [=Completion Record=] to model exception-based control
-flow. Because <a href=#algorithm-control-flow>web specifications have first-class support for exception handling</a>,
-special care must be taken when interpreting the results of ECMA-262 abstract operations.
-[[!ECMA-262]]
+<p>ECMA-262 defines some algorithms as [=abstract operations=] which may be invoked from algorithms
+in other specifications. Some of these [=abstract operations=] return a type of value named
+[=Completion Record=], and ECMA262's calling conventions use this type to model control flow
```
jmdyck, line 654:
```suggestion
[=Completion Record=], which ECMA-262 uses to model control flow
```

infra.bs
```diff
@@ -648,14 +649,20 @@ Standard that should be reported and addressed.
...
+interruptions such as thrown exceptions. Because web specifications do not observe the same
+conventions (e.g. <a href=#algorithm-control-flow>exception handling is built in to the processing
+model</a>), preserving the operations' semantics necessarily involves interpreting [=Completion
```
jmdyck, line 657: "interpreting" doesn't sound right to me. Maybe "handling"?

infra.bs
```diff
@@ -648,14 +649,20 @@ Standard that should be reported and addressed.
...
+Records=]. [[!ECMA-262]]
+
+<p>This specification defines a shorthand named <dfn export lt=throw-if-abrupt>?</dfn> as a
+convenience for web specifications wishing to faithfully interpret ECMA-262 [=abstract operations=]
+which return [=Completion Records=]. Algorithms which invoke such operations are encouraged to
+integrate them them concisely and consistently using this shorthand.
```
jmdyck, line 663:
```suggestion
integrate them concisely and consistently using this shorthand.
```

jmdyck:
> Those details don't seem necessary to explain the motivation or usage of this shorthand, though, so I've true to accommodate them simply by writing in less sweeping terms ("to model control flow interruptions such as thrown exceptions.")

Sounds good.

infra.bs
jugglinmike: You caught me being wordy--thanks

infra.bs
jugglinmike: Sure!

jugglinmike:
@annevk I think this is ready for another look!

annevk (commented):
How does this end up being different from JS's ? shorthand? This definition doesn't actually define anything as far as I can tell.

jmdyck:
The difference is in the way that an abrupt completion is handled. In ECMAScript, it's simply returned (i.e., the algorithm using `?` returns an abrupt completion), whereas here, its `[[Value]]` is thrown (i.e., the algorithm using `?` throws an Infra-exception). Issue #518 explains the significance of that distinction.

annevk:
Thanks for the refresher. So I think what needs to happen is that we state that normatively. And then have an example that illustrates it. And not have the normative requirements be made through examples.

Shawna1987 (commented):
Approve

Shawna1987 (commented):
approve

infra.bs
```diff
@@ -644,6 +647,46 @@ Standard that should be reported and addressed.
...
+  <ol>
+    <li><p>Let |result| be <a lt=throw-if-abrupt>?</a> AbstractOperation().
+  </ol>
+</div>
+
+<p>mean the same thing as:
+
+<div class=example id=example-throw-if-abrupt-expanded>
+  <ol>
+    <li><p>Let |hygienicTemp| be AbstractOperation().
```
Frankin45: > 
```suggestion
    <li><p>Let |hygienicTemp| be ### **AbstractOperation().**
```

PR #581, opened 2023-05-25
Algorithms should specify their return type in more cases.

jyasskin:
It's wrong to assert that return types are always easy to infer from
algorithm steps and call sites. We also shouldn't make people read
through the call sites (which might be in other documents) to figure out
what an algorithm returns.

***
<a href="https://whatpr.org/infra/581.html#algorithm-declaration" title="Last updated on May 25, 2023, 8:42 PM UTC (6d395a0)">Preview</a> | <a href="https://whatpr.org/infra/581/0dba3a2...6d395a0.html" title="Last updated on May 25, 2023, 8:42 PM UTC (6d395a0)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -297,8 +297,8 @@ a [=byte sequence=] <var ignore>bytes</var>, perform the following steps. They r
 [=string=] or null.
 
 <p>Algorithms which do not return a value use a shorter form. This same shorter form can be used even
-for algorithms that do return a value, since the return type is relatively easy to infer from the
-algorithm steps and call sites:
+for algorithms that do return a value if the return type is relatively easy to infer from the
+algorithm steps:
 
 <p class=exemplary-prose>To <dfn ignore>[algorithm name]</dfn>, given a [type1]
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -297,8 +297,8 @@ a [=byte sequence=] <var ignore>bytes</var>, perform the following steps. They r
 [=string=] or null.
 
 <p>Algorithms which do not return a value use a shorter form. This same shorter form can be used even
-for algorithms that do return a value, since the return type is relatively easy to infer from the
-algorithm steps and call sites:
+for algorithms that do return a value if the return type is relatively easy to infer from the
+algorithm steps:
```
annevk, line 301: Comma still seems good? Not entirely sure, but it's a pretty long sentence otherwise.

I was also thinking that we could say "nearby callers", but if you have to look at callers at all it's probably best clarified anyway.

jyasskin: I removed the comma because before it was ok to skip the second half of the sentence, but now it's not ok. But I don't care very much. :)

domenic (approved):

PR #589, opened 2023-06-26
Add a definition of less-than sign

shhnjk:
Define less-than sign which needs to be used in https://github.com/whatwg/html/pull/9309.

***
<a href="https://whatpr.org/infra/589.html" title="Last updated on Jun 26, 2023, 4:13 AM UTC (afa01db)">Preview</a> | <a href="https://whatpr.org/infra/589/dbc268d...afa01db.html" title="Last updated on Jun 26, 2023, 4:13 AM UTC (afa01db)">Diff</a>


domenic:
I don't think this is necessary; just use `U+003C (<)`.

annevk:
Indeed, we don't write out the name for characters that can literally be shown.

PR #635, opened 2024-04-15
Define isomorphic string

annevk:
And define it alongside ASCII and scalar value strings.

---

I think this will be useful for Cookies, but I need to think about it some more. But also in general we should probably be able to assert around isomorphic decode/encode?

***
<a href="https://whatpr.org/infra/635.html" title="Last updated on Nov 6, 2024, 4:06 PM UTC (4b9d443)">Preview</a> | <a href="https://whatpr.org/infra/635/4db814c...4b9d443.html" title="Last updated on Nov 6, 2024, 4:06 PM UTC (4b9d443)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -827,11 +827,13 @@ following steps return true:
 
 <hr>
 
+<div algorithm>
 <p>To <dfn export>isomorphic decode</dfn> a <a>byte sequence</a> <var>input</var>, return a
 <a>string</a> whose <a for=string>code point length</a> is equal to <var>input</var>'s
 <a for="byte sequence">length</a> and whose <a>code points</a> have the same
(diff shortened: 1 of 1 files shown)


annevk:
I looked at this again and given that it gives us a type we don't even need the assert anymore. I've implemented this inline where applicable. I think that definitely makes this a valuable addition.

domenic (approved):
Nice, I agree that in the context of already having "scalar value string", this is a win.

Be sure to update the commit message to also include ASCII string.

annevk:
ASCII string is already defined. This just moves the definition as is noted in the commit message. Unless I misunderstood?

domenic:
You're right, I just missed it.

Issue #509, opened 2022-12-13
Multiple assignment list unwrapping


dlrobertson:
## Summary

Similar to https://github.com/whatwg/infra/issues/508, it would be nice to be able to assign multiple variables from a [list](https://infra.spec.whatwg.org/#list) in a single unwrapping statement.

## Example

[strictly split](https://infra.spec.whatwg.org/#strictly-split) returns a [list](https://infra.spec.whatwg.org/#list), but in cases like [validate and extract](https://dom.spec.whatwg.org/#validate-and-extract), populating local variables with indexing leads to text that is a bit verbose. We currently have something like the following steps:

  1. Let _splitResult_ be the result of running [strictly split](https://infra.spec.whatwg.org/#strictly-split) given _qualifiedName_ and U+003A (:).
  2. Set _prefix_ to _splitResult_[0].
  3. Set _localName_ to _splitResult_[1].

With something like https://github.com/whatwg/infra/issues/508 for [list](https://infra.spec.whatwg.org/#list) we could simplify these three steps into something like the following:

  1. Set « _prefix_, _localName_ » to the result of running [strictly split](https://infra.spec.whatwg.org/#strictly-split) given _qualifiedName_ and U+003A (:).

## Context

_Originally posted by @annevk in https://github.com/whatwg/infra/issues/508#issuecomment-1348779181_

> I think in particular for strictly split it would be helpful if this worked on list-likes as well (the let-side could still use the tuple syntax I suppose), but maybe we ought to tackle that separately.

Also related to https://github.com/whatwg/dom/issues/1134.

inexorabletash:
This comes up extremely frequently in [WebNN](https://webmachinelearning.github.io/webnn/)

* Parameters like size (width, height), padding (top/bottom/left/right), etc. passed as lists
* A layout parameter (e.g. "nchw") guides the unpacking of a list of the dimensions of an input tensor (multidimensional array) into 4 different variables (batches, channels, height, width) - this looks weird at first but this is an industry-standard thing

I agree that a variation of the tuple syntax would be good. And I like @dlrobertson's suggestion that the LHS should use « » to be a cue that we're talking about a list on the RHS, i.e.:

> 1. Let « _batchSize_, _inputSize_ » be _inputShape_.

[...]

PR #638, opened 2024-08-11
Specify the convenience method for List multiple assignment

dlrobertson:
Add documentation and a example to specify and document List mulitple assignment.

Fixes: #509 

CC: @inexorabletash

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -1426,6 +1426,28 @@ out-of-bounds, except when used with <a for=list>exists</a>.
 "<code>b</code>", "<code>c</code>", "<code>a</code>" ». Then |example|[1] is the <a>string</a>
 "<code>b</code>".
 
+<p>For notational convenience, a multiple assignment syntax may be use to assign multiple variables
+to the <a>list</a>'s <a for=list>items</a>, by surrounding the variables to be assigned by « »
+characters and separating each variable name with a comma. To
+<dfn export for=list>assign multiple variables</dfn> from a <a>list</a> is to <a>Assert</a> that the
+<a>list</a>'s <a for=list>size</a> is the same as the number of variables to be assigned, and set
+each variable given to the value of the <a>list</a>'s <a for=list>item</a> at the corresponding
+index.
+
+<div class=example id=example-list-multiple-assignment>
+ <ol>
+  <li><p> Let |value| be the <a>list</a> « "<code>a</code>", "<code>b</code>", "<code>c</code>" ».
+
+  <li><p> Let « |a|, |b|, |c| » be |value|.
+
+  <li><p> <a>Assert</a>: |a| is "<code>a</code>"
+
+  <li><p> <a>Assert</a>: |b| is "<code>b</code>"
+
+  <li><p> <a>Assert</a>: |c| is "<code>c</code>"
+ </ol>
+</div>
+
 <hr>
 
 <p>To <dfn export for=list>append</dfn> to a <a>list</a> that is not an <a>ordered set</a> is to


dlrobertson:
First pass at adding multiple assignment for lists. Comments and critiques are very much welcomed

infra.bs
```diff
@@ -1426,6 +1426,28 @@ out-of-bounds, except when used with <a for=list>exists</a>.
 "<code>b</code>", "<code>c</code>", "<code>a</code>" ». Then |example|[1] is the <a>string</a>
 "<code>b</code>".
 
+<p>For notational convenience, a multiple assignment syntax may be use to assign multiple variables
+to the <a>list</a>'s <a for=list>items</a>, by surrounding the variables to be assigned by « »
+characters and separating each variable name with a comma. To
+<dfn export for=list>assign multiple variables</dfn> from a <a>list</a> is to <a>Assert</a> that the
```
annevk, line 1432: I don't think we need a definition as we'd never reference it.

I also don't like the way this "Assert" is inline and with an uppercase "a". Hmm. I think as above ("The index cannot be ...") we can state it as a requirement instead.

infra.bs
```diff
@@ -1426,6 +1426,28 @@ out-of-bounds, except when used with <a for=list>exists</a>.
...
+<a>list</a>'s <a for=list>size</a> is the same as the number of variables to be assigned, and set
+each variable given to the value of the <a>list</a>'s <a for=list>item</a> at the corresponding
+index.
+
+<div class=example id=example-list-multiple-assignment>
+ <ol>
+  <li><p> Let |value| be the <a>list</a> « "<code>a</code>", "<code>b</code>", "<code>c</code>" ».
```
annevk, line 1439: No need for a space to follow `<p>` here and below.

annevk (commented):
Nice, thanks for tackling this!

annevk:
Should we also add an example where we have a size check before assignment?

infra.bs
```diff
@@ -1426,6 +1426,36 @@ out-of-bounds, except when used with <a for=list>exists</a>.
...
+
+<p class=note>It is recommended to check that the <a>list</a>'s <a for=list>size</a> is the
+expected length before using the list multiple assignment syntax.
+
+<div class=example id=example-list-multiple-assignment-size>
+ <ol>
+  <li><p><a>Assert</a>: <a>list</a>'s <a for=list>size</a> is <code>3</code>.
+  <li><p>Let « |a|, |b|, |c| » be <a>list</a>.
+ </ol>
+</div>
```
dlrobertson, line 1457: > Should we also add an example where we have a size check before assignment?

Yeah, I think so. This is my first attempt at an example, but I'm not exactly sure if this is the best way to go about it is.

More contex form #509:

> I think I would prefer a two-step algorithm over handling a size mismatch inline, though I cannot really fault the English.

Does the added not and example sufficiently emphasize this?

infra.bs
```diff
@@ -1426,6 +1426,36 @@ out-of-bounds, except when used with <a for=list>exists</a>.
 "<code>b</code>", "<code>c</code>", "<code>a</code>" ». Then |example|[1] is the <a>string</a>
 "<code>b</code>".
 
+<p>For notational convenience, a multiple assignment syntax may be use to assign multiple variables
```
annevk, line 1429:
```suggestion
<p>For notational convenience, a multiple assignment syntax may be used to assign multiple variables
```

infra.bs
```diff
@@ -1426,6 +1426,36 @@ out-of-bounds, except when used with <a for=list>exists</a>.
...
+
+  <li><p><a>Assert</a>: |a| is "<code>a</code>"
+
+  <li><p><a>Assert</a>: |b| is "<code>b</code>"
+
+  <li><p><a>Assert</a>: |c| is "<code>c</code>"
+ </ol>
+</div>
+
+<p class=note>It is recommended to check that the <a>list</a>'s <a for=list>size</a> is the
```
annevk, line 1449: "recommended" is not a word you can use in a note: https://infra.spec.whatwg.org/#conformance

And this only really applies to lists you don't fully control, such as when they come from user input.

infra.bs
```diff
@@ -1426,6 +1426,36 @@ out-of-bounds, except when used with <a for=list>exists</a>.
...
+expected length before using the list multiple assignment syntax.
+
+<div class=example id=example-list-multiple-assignment-size>
+ <ol>
+  <li><p><a>Assert</a>: <a>list</a>'s <a for=list>size</a> is <code>3</code>.
+  <li><p>Let « |a|, |b|, |c| » be <a>list</a>.
```
annevk, line 1455: You need to make "list" a variable. I also think it would be slightly stronger to have the first line be something like:

1. If _list_'s size is not 3, then return failure.

as the size assert is already implicit in using the syntax.

infra.bs
```diff
@@ -1426,6 +1426,37 @@ out-of-bounds, except when used with <a for=list>exists</a>.
...
+</div>
+
+<p>When a <a>list</a>'s contents are not fully controlled, as s the case for lists from user input,
+the <a>list</a>'s <a for=list>size</a> should be checked to ensure it is the expected length before
+list multiple assignment syntax is used.
+
+<div class=example id=example-list-multiple-assignment-size>
+ <ol>
+  <li><p>If |list|'s <a for=list>size</a> is not <code>3</code>, then return failure.
+  <li><p>Let « |a|, |b|, |c| » be |value|.
```
annevk, line 1456:
```suggestion
  <li><p>Let « |a|, |b|, |c| » be |list|.
```

infra.bs
```diff
@@ -1426,6 +1426,37 @@ out-of-bounds, except when used with <a for=list>exists</a>.
...
+  <li><p><a>Assert</a>: |a| is "<code>a</code>"
+
+  <li><p><a>Assert</a>: |b| is "<code>b</code>"
+
+  <li><p><a>Assert</a>: |c| is "<code>c</code>"
+ </ol>
+</div>
+
+<p>When a <a>list</a>'s contents are not fully controlled, as s the case for lists from user input,
+the <a>list</a>'s <a for=list>size</a> should be checked to ensure it is the expected length before
```
annevk, line 1450:
```suggestion
the <a>list</a>'s <a for=list>size</a> should be checked to ensure it is the expected size before
```

annevk (commented):
@domenic thoughts?

I wonder if we should allow this same syntax to be used for splitting up a tuple. At least it was argued in the issue that using parenthesis here might not be clear enough and there's no need for disambiguation anyway.

infra.bs
```diff
@@ -1426,6 +1426,37 @@ out-of-bounds, except when used with <a for=list>exists</a>.
...
+
+  <li><p><a>Assert</a>: |a| is "<code>a</code>"
+
+  <li><p><a>Assert</a>: |b| is "<code>b</code>"
+
+  <li><p><a>Assert</a>: |c| is "<code>c</code>"
+ </ol>
+</div>
+
+<p>When a <a>list</a>'s contents are not fully controlled, as s the case for lists from user input,
```
domenic, line 1449:
```suggestion
<p>When a <a>list</a>'s contents are not fully controlled, as is the case for lists from user input,
```

domenic (approved):
LGTM with nits. Thanks for adding this!

infra.bs
```diff
@@ -1426,6 +1426,37 @@ out-of-bounds, except when used with <a for=list>exists</a>.
...
+ <ol>
+  <li><p>Let |value| be the <a>list</a> « "<code>a</code>", "<code>b</code>", "<code>c</code>" ».
+
+  <li><p>Let « |a|, |b|, |c| » be |value|.
+
+  <li><p><a>Assert</a>: |a| is "<code>a</code>"
+
+  <li><p><a>Assert</a>: |b| is "<code>b</code>"
+
+  <li><p><a>Assert</a>: |c| is "<code>c</code>"
```
domenic, line 1445:
```suggestion
  <li><p><a>Assert</a>: |a| is "<code>a</code>".

  <li><p><a>Assert</a>: |b| is "<code>b</code>".

  <li><p><a>Assert</a>: |c| is "<code>c</code>".
```

PR #648, opened 2024-11-15
Define Web IDL integer types and 128-bit unsigned

### LINKED ISSUE #647: Define ranged integer types
Although tackling #87 is a pretty big endeavor, I wonder if we can start smaller, by defining things such as:

- A **16-bit unsigned integer** is an integer in the range 0 to 65535, inclusive.

This could be shared across Web IDL, URL, and Declarative Web Push. And probably other specifications.

annevk:
Fixes #647.

***
<a href="https://whatpr.org/infra/648.html" title="Last updated on Nov 18, 2024, 7:57 AM UTC (d205621)">Preview</a> | <a href="https://whatpr.org/infra/648/79b945d...d205621.html" title="Last updated on Nov 18, 2024, 7:57 AM UTC (d205621)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -723,6 +723,47 @@ JavaScript <b>null</b> value. [[!ECMA-262]]
 <p class=example id=example-boolean>Let <var ignore>elementSeen</var> be false.
 
 
+<h3 id=numbers>Numbers</h3>
+
+<p class=XXX>Numbers are complicated, please see
+<a href="https://github.com/whatwg/infra/issues/87">issue #87</a>. In due course we hope to offer
+more guidance here around types and mathematical operations. Help appreciated!
+
+<hr>
+
+<p>An <dfn export>8-bit unsigned integer</dfn> is an integer in the range 0 to 255 (0 to
+2<sup>8</sup> &minus; 1), inclusive.
+
+<p>A <dfn export>16-bit unsigned integer</dfn> is an integer in the range 0 to 65535 (0 to
+2<sup>16</sup> &minus; 1), inclusive.
+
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -723,6 +723,47 @@ JavaScript <b>null</b> value. [[!ECMA-262]]
 <p class=example id=example-boolean>Let <var ignore>elementSeen</var> be false.
 
 
+<h3 id=numbers>Numbers</h3>
+
+<p class=XXX>Numbers are complicated, please see
```
domenic, line 728:
```suggestion
<p class=XXX>Numbers are complicated; please see
```

infra.bs
```diff
@@ -723,6 +723,47 @@ JavaScript <b>null</b> value. [[!ECMA-262]]
...
+
+<p>A <dfn export>128-bit unsigned integer</dfn> is an integer in the range 0 to
+340282366920938463463374607431768211455 (0 to 2<sup>128</sup> &minus; 1), inclusive.
+
+<p class=example id=example-128-bit-unsigned-integer>An IPv6 address is an
+<a for=/>128-bit unsigned integer</a>.
+
+<hr>
+
+<p>An <dfn export>8-bit signed integer</dfn> is an integer in the range -128 to 127 (-2<sup>7</sup>
```
domenic, line 754: In HTML at least, we use `&minus;` for negative signs too. I've always been a bit conflicted about that as I think it's typographically more correct, but it makes copying and pasting into code worse... I'm OK to leave it up to you but maybe we should raise a meta issue.

domenic (approved):
LGTM with nits

infra.bs
annevk: I'm happy to stick with `&minus;`. I should have double checked this in HTML.

PR #659, opened 2024-12-18
Define an algorithm to create a Set from a List

OrKoN:
In the WebDriver BiDi spec, there is a need to convert lists created from JSON messages into sets for internal algorithms. This PR adds a helper to define how the operation of creating a set from a list should be performed. The operation seems to be common and relatively straight-forward so it might make sense that the Infra standard defines it.

***
<a href="https://whatpr.org/infra/659.html" title="Last updated on Dec 19, 2024, 7:18 AM UTC (ca83296)">Preview</a> | <a href="https://whatpr.org/infra/659/7351436...ca83296.html" title="Last updated on Dec 19, 2024, 7:18 AM UTC (ca83296)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -1694,6 +1694,16 @@ unordered one, since interoperability requires that any developer-exposed enumer
 contents be consistent between browsers. In those cases where order is not important, we still use
 ordered sets; implementations can optimize based on the fact that the order is not observable.
 
+<div algorithm>
+<p>To <dfn export for=set>create</dfn> a [=/set=], from a [=/list=] |input|:
+
+<ol>
+ <li><p>Let |result| be an empty [=/set=].
+ <li><p>For each [=list/item=] |item| in |input|, [=set/append=] |item| to |result|.
+ <li><p>Return |result|.
+</ol>
+</div>
+
 <p>To <dfn export for=set>append</dfn> to an <a>ordered set</a>: if the set <a for=list>contains</a>
 the given <a for=set>item</a>, then do nothing; otherwise, perform the normal <a>list</a>
 <a for=list>append</a> operation.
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -1694,6 +1694,16 @@ unordered one, since interoperability requires that any developer-exposed enumer
 contents be consistent between browsers. In those cases where order is not important, we still use
 ordered sets; implementations can optimize based on the fact that the order is not observable.
 
+<div algorithm>
+<p>To <dfn export for=set>create a set</dfn>, given a [=/list=] |input|:
```
domenic, line 1698: Hmm, I'd prefer naming this either `<dfn export>create a set from a list</dfn>`, or `<dfn export for="set">create</dfn>`. The latter makes sense if we think there's no other reasonable input for creating a set, whereas the former is more specific.

domenic (commented):
Good idea!

infra.bs
```diff
@@ -1694,6 +1694,16 @@ unordered one, since interoperability requires that any developer-exposed enumer
...
+
+<ol>
+ <li><p>Let |result| be an empty [=/set=].
+ <li><p>For each [=list/item=] |item| in |input|, [=set/append=] |item| to |result|.
```
annevk, line 1702: for each x of y*

infra.bs
OrKoN: Went for `<dfn export for="set">create</dfn>`. I think creating from other data structures is less common. Nevertheless, the create steps could accept a generic iterator instead of a list but since that is not currently specified it probably makes sense to just use the list.

infra.bs
OrKoN: Thanks, changed to `for each x of y`, also linked `for each`.

annevk (approved):
I suppose we can always overload or chase down all callers using dontcallmedom's website in the worst case.

annevk:
Thanks Alex!

OrKoN:
Thanks for reviewing!

Issue #643, opened 2024-09-18
Equality of structs/tuples is undefined


tabatkins:
### What is the issue with the Infra Standard?

It's not currently specified how struct/tuple equality works. This is required, for example, for it to be well-defined how they work as map keys (and this is already done, see <https://html.spec.whatwg.org/multipage/webappapis.html#module-map>).

I think the obvious answer (and definitely what's expected for the module map) is that struct/tuple equality is structural - two structs/tuples are equivalent if they have the same shape, and the contents are equal.

domenic:
Related (but different): https://github.com/whatwg/html/issues/10315

annevk:
See also #230.

Defining this at the Infra level would require some kind of operator overloading and concepts that want to participate to consistently use "equals" to define their equality and such. I think it's something we have to do eventually, but it's quite a lift.

PR #660, opened 2025-01-10
Define equality for sets

OrKoN:
A definition of what it means for two sets to be equal. This definition re-uses the subset/superset definitions but we could also define it as an iteration over elements and size equality.

***
<a href="https://whatpr.org/infra/660.html" title="Last updated on Jan 10, 2025, 9:35 AM UTC (1aca0e5)">Preview</a> | <a href="https://whatpr.org/infra/660/7351436...1aca0e5.html" title="Last updated on Jan 10, 2025, 9:35 AM UTC (1aca0e5)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -1718,6 +1718,9 @@ of either with <var>replacement</var> and <a for=set>remove</a> all other instan
 <p class=note>This implies that an <a>ordered set</a> is both a <a for=set>subset</a> and a
 <a for=set>superset</a> of itself.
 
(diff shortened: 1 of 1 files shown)


sadym-chromium (commented):
SGTM

annevk:
I guess this works, but we're also running into #643 with these.

@domenic you okay with generally using "equal(s)" for structural equality and "is" for identity equality?

domenic (approved):
Yeah, this seems reasonable to me. Although it's a bit silly that we haven't defined what it means for two strings to be equal, or two numbers, or similar, so this is all on slightly shaky foundations.

PR #669, opened 2025-05-13
Get a map value with a default

### LINKED ISSUE #668: Get a map value with a default.
### What problem are you trying to solve?

Over at [sanitizer-api](https://github.com/WICG/sanitizer-api/), we deal a lot with user-supplied maps where a given key may be absent. [Get the value of an entry](https://infra.spec.whatwg.org/#map-get) asserts that a key is present, which requires to first check for the key presence. This gets quite "wordy" quickly, especially if accesses are chained.

### What solutions exist today?

Today, one can always add an if-then:

* If *a* *contains* *b*: 
  * ... do stuff with *a[b]* ...
  * Otherwise, handle default case.

Chained case:
* If *map* *contains* *a*:
  *  If *map[a]* *contains* *b*:
      * ... do stuff with *map[a][b]*.
      * Otherwise, return *default*.
  * Otherwise, return *default*.

### How would you solve it?

In https://github.com/WICG/sanitizer-api/pull/285, I added a definition ["maybe get the value of an entry"](https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/285.html#map-maybe-get-the-value), which basically extends infra's *get the value of an entry* with a default parameter, which it will return if a key isn't set. Plus a two-parameter shorthand, *map[key,default]*.

Examples:
- Safely accessing a map: *map["a", None]*
- Safely accessing a map of maps: *map["a",«[]»]["b", default]*

Definition plus usage examples:
https://pr-preview.s3.amazonaws.com/otherdaniel/purification/pull/285.html#map-maybe-get-the-value

Previous versions of the same PR use the existing syntax with if-guards.

### Anything else?

_No response_

otherdaniel:
This introduces an optional default argument to [=map/get the value=]. Also adds [=map/with default=] to enable this default with the indexing syntax.

Fixes #668.

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -1785,20 +1785,28 @@ entries with a comma.
 <hr>
 
 <p>To <dfn export for=map lt="get|get the value">get the value of an entry</dfn> in an
-<a>ordered map</a> <var>map</var> given a <a for=map>key</a> <var>key</var>:
+<a>ordered map</a> <var>map</var> given a <a for=map>key</a> <var>key</var> and an optional
+<var>default</var>:
 
 <ol>
+ <li><p>If <var>map</var> does not <a for=map>contain</a> <var>key</var> and <var>default</var>
+ is given, then return <var>default</var>.
  <li><p>Assert: <var>map</var> <a for=map>contains</a> <var>key</var>.
-
  <li><p>Return the <a for=map>value</a> of the <a for=map>entry</a> in <var>map</var> whose
  <a for=map>key</a> is <var>key</var>.
 </ol>
 
 <p>We can also denote <a for=map lt=get>getting the value of an entry</a> using an indexing syntax,
-by providing a <a for=map>key</a> inside square brackets directly following a <a for=/>map</a>.
+by providing a <a for=map>key</a> inside square brackets directly following a <a for=/>map</a>. A default can be given by adding the phrase <dfn export for=map>with default</dfn> followed by the default value.
 
 <p class=example id=example-map-get>If <var ignore>map</var>["<code>test</code>"]
 <a for=map>exists</a>, then return <var ignore>map</var>["<code>test</code>"].
+
+<p class=example id=example-map-get-with-default>Let <var>example</var> be the <a>ordered map</a>
+«[ "<code>a</code>" → "<code>x</code>", "<code>b</code>" → "<code>y</code>" ]». Then
+|example|["<code>a</code>"] is the same as |example|["<code>a</code>"] <a>with default</a> "<code>z</code>",
+namely "<code>x</code>". |example|["<code>c</code>"] is invalid.
+|example|["<code>c</code>"] <a>with default</a> "<code>z</code>" is "<code>z</code>".</p>
 
 <p>To <dfn export for=map lt="set|set the value">set the value of an entry</dfn> in an
 <a>ordered map</a> to a given <a for=map>value</a> is to update the value of any existing


infra.bs
```diff
@@ -1785,21 +1785,29 @@ entries with a comma.
 <hr>
 
 <p>To <dfn export for=map lt="get|get the value">get the value of an entry</dfn> in an
-<a>ordered map</a> <var>map</var> given a <a for=map>key</a> <var>key</var>:
+<a>ordered map</a> <var>map</var> given a <a for=map>key</a> <var>key</var> and an optional
+<var>default</var>:
```
annevk, line 1789: This is missing a type declaration for _default_.

infra.bs
```diff
@@ -1785,21 +1785,29 @@ entries with a comma.
...
 
 <ol>
+ <li><p>If <var>map</var> does not <a for=map>contain</a> <var>key</var> and <var>default</var>
+ is given, then return <var>default</var>.
```
annevk, line 1793: Newline after this step.

infra.bs
```diff
@@ -1785,21 +1785,29 @@ entries with a comma.
...
  <li><p>Assert: <var>map</var> <a for=map>contains</a> <var>key</var>.
-
```
annevk, line 1792: Please don't remove this.

infra.bs
```diff
@@ -1785,21 +1785,29 @@ entries with a comma.
...
-by providing a <a for=map>key</a> inside square brackets directly following a <a for=/>map</a>.
+by providing a <a for=map>key</a> inside square brackets directly following a <a for=/>map</a>. A default can be given by adding the phrase <dfn export for=map>with default</dfn> followed by the default value.
 
 <p class=example id=example-map-get>If <var ignore>map</var>["<code>test</code>"]
 <a for=map>exists</a>, then return <var ignore>map</var>["<code>test</code>"].
 
+<p class=example id=example-map-get-with-default>Let <var>example</var> be the <a>ordered map</a>
+«[ "<code>a</code>" → "<code>x</code>", "<code>b</code>" → "<code>y</code>" ]». Then
+|example|["<code>a</code>"] is the same as |example|["<code>a</code>"] <a>with default</a> "<code>z</code>",
+namely "<code>x</code>". |example|["<code>c</code>"] is invalid.
```
annevk, line 1808:
```suggestion
namely "<code>x</code>". |example|["<code>c</code>"] would hit an assert.
```

infra.bs
```diff
@@ -1785,21 +1785,29 @@ entries with a comma.
...
+ is given, then return <var>default</var>.
  <li><p>Assert: <var>map</var> <a for=map>contains</a> <var>key</var>.
-
  <li><p>Return the <a for=map>value</a> of the <a for=map>entry</a> in <var>map</var> whose
  <a for=map>key</a> is <var>key</var>.
 </ol>
 
 <p>We can also denote <a for=map lt=get>getting the value of an entry</a> using an indexing syntax,
-by providing a <a for=map>key</a> inside square brackets directly following a <a for=/>map</a>.
+by providing a <a for=map>key</a> inside square brackets directly following a <a for=/>map</a>. A default can be given by adding the phrase <dfn export for=map>with default</dfn> followed by the default value.
```
annevk, line 1800: We wrap at 100 columns.

annevk (commented):
This looks pretty good to me.

infra.bs
domenic: I don't think we necessarily need one, since it can be any value.

infra.bs
```diff
@@ -1785,21 +1785,29 @@ entries with a comma.
...
 
 <p class=example id=example-map-get>If <var ignore>map</var>["<code>test</code>"]
 <a for=map>exists</a>, then return <var ignore>map</var>["<code>test</code>"].
 
+<p class=example id=example-map-get-with-default>Let <var>example</var> be the <a>ordered map</a>
+«[ "<code>a</code>" → "<code>x</code>", "<code>b</code>" → "<code>y</code>" ]». Then
+|example|["<code>a</code>"] is the same as |example|["<code>a</code>"] <a>with default</a> "<code>z</code>",
+namely "<code>x</code>". |example|["<code>c</code>"] is invalid.
+|example|["<code>c</code>"] <a>with default</a> "<code>z</code>" is "<code>z</code>".</p>
```
domenic, line 1809:
```suggestion
|example|["<code>c</code>"] <a>with default</a> "<code>z</code>" is "<code>z</code>".
```

domenic (approved):
LGTM with nits

infra.bs
otherdaniel: Done.

infra.bs
otherdaniel: Reinstanted.

infra.bs
otherdaniel: Done.

infra.bs
otherdaniel: Not sure if this is what is meant, but I've now declared it as a map/value, since that corresponds to the role that |default| takes on here.

annevk: I guess that works, although those are not really types and maybe you want something outside of what the tuple can contain. Hmm.

I briefly thought of defining "any/anything" but it will look so ugly in prose. Slowly coming around to @domenic's side of leaving the type out when it can be anything, even if it's a bit ambiguous. @domenic any final thoughts?

domenic: I think leaving the type out is best.

otherdaniel: I removed the type again.

annevk (approved):

annevk:
Thanks @otherdaniel!

Issue #686, opened 2025-11-19
Consider supporting named algorithm return values.


mikewest:
### What problem are you trying to solve?

As discussed in https://github.com/whatwg/fetch/pull/1867#discussion_r2538122239, it's sometimes helpful (and at least somewhat common practice) to return some prose value from an algorithm that makes the result clear. For example:

* https://fetch.spec.whatwg.org/#cross-origin-resource-policy-check returns "blocked" or "allowed" (as do several algorithms in CSP, SRI, etc).
* Several algorithms in URL, HTML, etc return "failure" (e.g. https://url.spec.whatwg.org/#concept-domain-to-ascii, https://url.spec.whatwg.org/#concept-host-parser, https://html.spec.whatwg.org/#rules-for-parsing-dimension-values, ...).
* WebDriver defines "success" and "error" (https://w3c.github.io/webdriver/#algorithms).

These aren't well-supported by Infra. Perhaps they should be?

### How would you solve it?

It seems ideal to agree on some way of representing these kinds of algorithms and their results in Infra.

Based on very little research, I'd suggest that we define "failure" in any event, and consider defining "boolean" algorithms in some way, with a set of truthy values ("Allowed", "Valid", "Success", whatever other common patterns we find) and a set of falsy values ("Blocked", "Invalid", "Failure", etc).

tabatkins:
@aphillips Note that this is talking about internal spec algorithm values, not author-exposed enumerations.

PR #687, opened 2025-11-20
Define singleton types.

mikewest:
This commit defines a set of "singleton" types which are commonly used as inputs or outputs to prose algorithms: "success"/"failure", and "allowed"/"blocked".

This is a first pass at whatwg/infra#686 for discussion.

I'm not entirely convinced about the name, but naming is hard and I'm happy to run with "singleton types" if that's what makes sense to folks. Other things that occurred to me were "[sentinel](https://en.wikipedia.org/wiki/Sentinel_value) types" or shifting this away from the "type" framing into the algorithm section as "named inputs/outputs". But I don't have strong opinions.

***
<a href="https://whatpr.org/infra/687.html" title="Last updated on Nov 24, 2025, 7:31 AM UTC (34ef09a)">Preview</a> | <a href="https://whatpr.org/infra/687/096add4...34ef09a.html" title="Last updated on Nov 24, 2025, 7:31 AM UTC (34ef09a)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -1484,7 +1484,7 @@ these steps:
 <h3 id=singleton>Singletons</h3>
 
 The following types can be used as inputs or outputs of algorithmic prose to improve comprehension.
-They have no web-visible representation beyond their effect on the flow of any given algorithm, and
+They have no observable representation beyond their effect on the flow of any given algorithm, and
 they can only be compared with "is".
 
 <dl>
@@ -1517,7 +1517,7 @@ they can only be compared with "is".
  <ol>
   <li><p>If <var>result</var> is <a>failure</a>, then shut everything down gracefully and abort.
   <li><p><a>Assert</a>: <var>result</var> is <a>success</a>.
-  <li><p>...
+  <li><p>…
  </ol>
 </div>
 


infra.bs
```diff
@@ -1481,6 +1481,47 @@ these steps:
...
+
+ <ol>
+  <li><p>If something unexpectedly bad happened, then return <a>failure</a>.
+  <li><p>Otherwise, return <a>success</a>.
+ </ol>
+
+ <p>Callers of that algorithm might evaluate the result as follows:
+
+ <ol>
+  <li><p>If <var>result</var> is <a>failure</a>, shut everything down gracefully and abort.
```
annevk:
```suggestion
  <li><p>If <var>result</var> is <a>failure</a>, then shut everything down gracefully and abort.
```

annevk (commented):
I would place the example after introducing the 4 singletons.

This looks okay to me. I do wonder if we want null in here as well as it's rather similar, but maybe keeping it separate for now is okay as it has some additional semantics.

mikewest:
> I do wonder if we want null in here as well as it's rather similar, but maybe keeping it separate for now is okay as it has some additional semantics.

Given that null maps to the JavaScript value, it made sense to me to leave it as a distinct thing. I'm happy to combine these all into one section, though, if that makes more sense to you.

PR #706, opened 2026-04-09
Add list slice

jakearchibald:
For https://github.com/whatwg/html/pull/12345.

I also changed the definition of clone, since it's just a slice with default args.

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -1662,13 +1662,17 @@ set of steps on each <a for=list>item</a> in order, use phrasing of the form
 subsequent prose.
 
 <div algorithm=slice>
-<p>To <dfn export for=list,stack,queue,set>slice</dfn> a <a>list</a> |list|, with an optional index
-<dfn export for=list/slice><var>from</var></dfn> (default 0) and an optional index
+<p>To <dfn export for=list,stack,queue,set>slice</dfn> a <a>list</a> |list|, with an optional
+integer <dfn export for=list/slice><var>from</var></dfn> (default 0) and an optional integer
 <dfn export for=list/slice><var>to</var></dfn> (default |list|'s <a for=list>size</a>), perform the
 following steps. They return a <a>list</a> of the same designation as |list|.
 
 <ol>
  <li><p><a>Assert</a>: |to| is greater than or equal to |from|.
+
+ <li><p>If |from| is less than 0, then set |from| to 0.
+
+ <li><p>If |to| is greater than |list|'s <a for=list>size</a>, then set |to| to |list|'s <a for=list>size</a>.
 
  <li><p>Let |slice| be a new <a>list</a> of the same designation as |list|.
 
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -1661,10 +1661,15 @@ set of steps on each <a for=list>item</a> in order, use phrasing of the form
...
-<p>To <dfn export for=list,stack,queue,set>clone</dfn> a <a>list</a> |list| is to create a new
-<a>list</a> |clone|, of the same designation, and, <a for=list>for each</a> |item| of |list|,
-<a for=list>append</a> |item| to |clone|, so that |clone| <a for=list>contains</a> the same
-<a for=list>items</a>, in the same order as |list|.
+<p>To <dfn export for=list,stack,queue,set>slice</dfn> a <a>list</a> |list|, with an optional index
+<dfn export for=list/slice><var>from</var></dfn> (default 0) and an optional index
+<dfn export for=list/slice><var>to</var></dfn> (default |list|'s <a for=list>size</a>), is to create a new
+<a>list</a> |slice|, of the same designation, and <a for=list>append</a> each
+<a for=list>item</a> of |list|, where the index is greater than or equal to |from| and less than
+|to|, to |slice|, maintaining order.
```
annevk: I think we should make this a list of steps for clarity. That way we can also assert the obvious things about _from_ and _to_.

jakearchibald: Yeah that's fair.

Are you ok with clone being defined in terms of slice?

annevk: Yeah that seems good. Always nice to have fewer fundamental primitives.

jakearchibald: Done!

infra.bs
```diff
@@ -1661,10 +1661,30 @@ set of steps on each <a for=list>item</a> in order, use phrasing of the form
...
+integer <dfn export for=list/slice><var>from</var></dfn> (default 0) and an optional integer
+<dfn export for=list/slice><var>to</var></dfn> (default |list|'s <a for=list>size</a>), perform the
+following steps. They return a <a>list</a> of the same designation as |list|.
+
+<ol>
+ <li><p><a>Assert</a>: |to| is greater than or equal to |from|.
+
+ <li><p>If |from| is less than 0, then set |from| to 0.
+
+ <li><p>If |to| is greater than |list|'s <a for=list>size</a>, then set |to| to |list|'s <a for=list>size</a>.
```
jakearchibald: I felt this was preferable to asserting. It means, like in JS, you can slice from 0 to 100 to 'crop' a list to maximum of 100 items.

```js
['foo'].slice(0, 100); // does not throw
```

annevk: Is this needed for specifications? I might not be considering all scenarios that well, but I was assuming we'd just want to require both of these to be within bounds of the list.

jakearchibald: I don't mind turning these into asserts. I was just copying how JS `array.slice()` works.

annevk: Let's start out with a very strict contract then until we hit cases where it would make sense to not be so strict.

jakearchibald: Done!

annevk (approved):

jakearchibald:
Cheers!

PR #709, opened 2026-05-06
Add `has duplicates` to list, and let `remove` return a boolean of whether items were removed

noamr:
These algorithms are currently defined in the sanitizer spec, and seem like they belong in infra instead.

***
<a href="https://whatpr.org/infra/709.html" title="Last updated on May 7, 2026, 11:25 AM UTC (6fc6344)">Preview</a> | <a href="https://whatpr.org/infra/709/994009c...6fc6344.html" title="Last updated on May 7, 2026, 11:25 AM UTC (6fc6344)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -1712,16 +1712,19 @@ To <dfn export for=list>remove duplicates</dfn> from a <a>list</a> <var>list</va
 </div>
 
 <div algorithm=has-duplicates>
-To check if a <a>list</a> |list| <dfn export for=list>has duplicates</dfn>, perform the following steps:
+To check if a <a>list</a> |list| <dfn export for=list>has duplicates</dfn>:
 
 <ol>
- <li><p>Let <var>result</var> be an empty <a>list</a>.
+ <li><p>Let <var>result</var> be « ».
 
- <li><p><a for=set>For each</a> |item| of <var>list</var>:
+ <li>
+  <p><a for=set>For each</a> |item| of <var>list</var>:</p>
 
   <ol>
    <li><p>If <var>result</var> <a for=list>contains</a> |item|, return true.</p></li>
+   <li><p><a for=list>Append</a> |item| to <var>result</var>.</p></li>
   </ol>
+ </li>
 
  <li><p>Return false.</p></li>
 </ol>
--- a/infra.bs
+++ b/infra.bs
@@ -1691,7 +1691,7 @@ To <dfn export for=list,set>potentialy remove</dfn> zero or more <a for=list>ite
 
  <li><p><a for=list,set>Remove</a> the items from |list| based on the given condition.
 
- <li><p>Return true if |list|'s <a for=list>size</a> is less than |sizeBefore|, and false otherwise.
+ <li><p>Return true if |list|'s <a for=list>size</a> is less than |sizeBefore|; otherwise false.
 </ol>
 </div>
 
--- a/infra.bs
+++ b/infra.bs
@@ -1692,7 +1692,7 @@ following steps. They return a <a>list</a> of the same designation as |list|.
 </ol>
 </div>
 
-<div algorithm=has-duplicates>
+<div algorithm>
 To check if a <a>list</a> |list| <dfn export for=list>has duplicates</dfn>:
 
 <ol>
--- a/infra.bs
+++ b/infra.bs
@@ -1702,7 +1702,7 @@ To check if a <a>list</a> |list| <dfn export for=list>has duplicates</dfn>:
   <p><a for=set>For each</a> |item| of <var>list</var>:
 
   <ol>
-   <li><p>If <var>seenItems</var> <a for=list>contains</a> |item|, return true.</p></li>
+   <li><p>If <var>seenItems</var> <a for=list>contains</a> |item|, then return true.</p></li>
 
    <li><p><a for=list>Append</a> |item| to <var>seenItems</var>.</p></li>
   </ol>
--- a/infra.bs
+++ b/infra.bs
@@ -1687,7 +1687,7 @@ following steps. They return a <a>list</a> of the same designation as |list|.
 To <dfn export for=list,set>potentialy remove</dfn> zero or more <a for=list>items</a> from a <a>list</a> |list| based on a condition:
 
 <ol>
- <li><p>If no item in |list| matches the given condition <a for=list,set>exists</a>, return false.
+ <li><p>If no item in |list| matches the given condition <a for=list,set>exists</a>, then return false.
 
  <li><p><a for=list,set>Remove</a> the items from |list|.
 
--- a/infra.bs
+++ b/infra.bs
@@ -1684,7 +1684,7 @@ following steps. They return a <a>list</a> of the same designation as |list|.
 </div>
 
 <div algorithm=porentially-remove>
-To <dfn export for=list,set>potentialy remove</dfn> zero or more <a for=list>items</a> from a <a>list</a> |list| based on a condition, perform the following steps:
+To <dfn export for=list,set>potentialy remove</dfn> zero or more <a for=list>items</a> from a <a>list</a> |list| based on a condition:
 
 <ol>
  <li><p>If no item in |list| matches the given condition <a for=list,set>exists</a>, return false.
--- a/infra.bs
+++ b/infra.bs
@@ -1695,22 +1695,6 @@ To <dfn export for=list,set>potentialy remove</dfn> zero or more <a for=list>ite
 </ol>
 </div>
 
-<div algorithm=remove-duplicates>
-To <dfn export for=list>remove duplicates</dfn> from a <a>list</a> <var>list</var>, perform the following steps:
-
-<ol>
- <li><p>Let <var>result</var> be an empty <a>list</a>.
-
- <li><p><a for=set>For each</a> |item| of <var>list</var>:
-
-  <ol>
-   <li><p>If <var>result</var> does not <a for=list>contain</a> |item|, <a for=list>append</a> |item| to <var>result</var>.</p></li>
-  </ol>
-
- <li><p>Return <var>result</var>.</p></li>
-</ol>
-</div>
-
 <div algorithm=has-duplicates>
 To check if a <a>list</a> |list| <dfn export for=list>has duplicates</dfn>:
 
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -1683,6 +1683,50 @@ following steps. They return a <a>list</a> of the same designation as |list|.
 </ol>
 </div>
 
+<div algorithm=porentially-remove>
+To <dfn export for=list,set>potentialy remove</dfn> zero or more <a for=list>items</a> from a <a>list</a> |list| based on a condition, perform the following steps:
+
+<ol>
+ <li><p>If no item in |list| matches the given condition <a for=list,set>exists</a>, return false.
```
annevk: then return*

Also, I guess this matches precedent and we haven't formalized condition elsewhere either?

infra.bs
```diff
@@ -1683,6 +1683,50 @@ following steps. They return a <a>list</a> of the same designation as |list|.
...
+
+ <li><p><a for=list,set>Remove</a> the items from |list|.
+
+ <li><p>Return true.
```
annevk: Should we really return true if the items are not in the set/list? Or is it a precondition that they are? If the latter, we should assert. If the former, I don't think I like it.

infra.bs
```diff
@@ -1683,6 +1683,50 @@ following steps. They return a <a>list</a> of the same designation as |list|.
...
+</ol>
+</div>
+
+<div algorithm=remove-duplicates>
+To <dfn export for=list>remove duplicates</dfn> from a <a>list</a> <var>list</var>, perform the following steps:
```
annevk: I think this is https://infra.spec.whatwg.org/#set-create

Or do you really still need a list in the end?

infra.bs
```diff
@@ -1683,6 +1683,50 @@ following steps. They return a <a>list</a> of the same designation as |list|.
...
+  <ol>
+   <li><p>If <var>result</var> does not <a for=list>contain</a> |item|, <a for=list>append</a> |item| to <var>result</var>.</p></li>
+  </ol>
+
+ <li><p>Return <var>result</var>.</p></li>
+</ol>
+</div>
+
+<div algorithm=has-duplicates>
+To check if a <a>list</a> |list| <dfn export for=list>has duplicates</dfn>, perform the following steps:
```
annevk:
```suggestion
To check if a <a>list</a> |list| <dfn export for=list>has duplicates</dfn>:
```

infra.bs
```diff
@@ -1683,6 +1683,50 @@ following steps. They return a <a>list</a> of the same designation as |list|.
...
+
+<ol>
+ <li><p>Let <var>result</var> be an empty <a>list</a>.
```
annevk: Use the syntax here.

infra.bs
```diff
@@ -1683,6 +1683,50 @@ following steps. They return a <a>list</a> of the same designation as |list|.
...
+
+ <li><p><a for=set>For each</a> |item| of <var>list</var>:
+
+  <ol>
+   <li><p>If <var>result</var> <a for=list>contains</a> |item|, return true.</p></li>
```
annevk: This can be inlined above after the for each colon with a lowercase "if". Also then return

infra.bs
```diff
@@ -1683,6 +1683,50 @@ following steps. They return a <a>list</a> of the same designation as |list|.
 </ol>
 </div>
 
+<div algorithm=porentially-remove>
+To <dfn export for=list,set>potentialy remove</dfn> zero or more <a for=list>items</a> from a <a>list</a> |list| based on a condition, perform the following steps:
```
annevk:
```suggestion
To <dfn export for=list,set>potentialy remove</dfn> zero or more <a for=list>items</a> from a <a>list</a> |list| based on a condition:
```

infra.bs
noamr: Methods like https://wicg.github.io/sanitizer-api/#dom-sanitizer-allowelement return a boolean indicating if the underlying lists was modified. So this functionality is necessary, but it can live in the call site (effectively moving "potentially remove" to the HTML spec).

infra.bs
noamr: Yes exactly

infra.bs
noamr: It's passed in as a dictionary member which might have duplicates. I don't know if we can implicitly treat it as a set

infra.bs
noamr: Actually it was missing a step.

infra.bs
noamr: Done

infra.bs
noamr: Done

infra.bs
annevk: I don't think you understood. Remove can no-op and you are not accounting for that.

infra.bs
annevk: I'm suggesting that the relevant algorithm could call set-create instead of remove-duplicates.

infra.bs
noamr: You're right, I didn't!
I refactored it to rely on list's size, it seems more readable maybe? Otherwise I can assert that the list actually removed something but this seems more normative

infra.bs
noamr: OK, we can do that sure. Removed "remove duplicates" from this PR

infra.bs
```diff
@@ -1683,6 +1683,38 @@ following steps. They return a <a>list</a> of the same designation as |list|.
...
 
+<div algorithm=porentially-remove>
+To <dfn export for=list,set>potentialy remove</dfn> zero or more <a for=list>items</a> from a <a>list</a> |list| based on a condition:
+
+<ol>
+ <li><p>Let |sizeBefore| be |list|'s <a for=list>size</a>.
+
+ <li><p><a for=list,set>Remove</a> the items from |list| based on the given condition.
+
+ <li><p>Return true if |list|'s <a for=list>size</a> is less than |sizeBefore|, and false otherwise.
```
annevk:
```suggestion
 <li><p>Return true if |list|'s <a for=list>size</a> is less than |sizeBefore|; otherwise false.
```

infra.bs
```diff
@@ -1683,6 +1683,38 @@ following steps. They return a <a>list</a> of the same designation as |list|.
...
+</ol>
+</div>
```
annevk: This looks good now. But I think we can fold this into "remove" and not add "potentially remove".

infra.bs
```diff
@@ -1683,6 +1683,38 @@ following steps. They return a <a>list</a> of the same designation as |list|.
...
+
+<div algorithm=has-duplicates>
+To check if a <a>list</a> |list| <dfn export for=list>has duplicates</dfn>:
+
+<ol>
+ <li><p>Let <var>result</var> be « ».
```
annevk: _result_ is a weird name as it's not the return value. Maybe _set_ or _seenItems_?

infra.bs
noamr: Done

infra.bs
noamr: Done

infra.bs
```diff
@@ -1626,8 +1626,17 @@ below for <a for=set lt=append>ordered set append</a>, <a for=set>prepend</a>, a
...
-to remove all items from the list that match a given condition, or do nothing if none do.
+<div algorithm=remove>
+<p>To <dfn export for=list,set>remove</dfn> zero or more <a for=list>items</a> from a <a>list</a>:
+
+<ol>
+<li><p>Let |sizeBefore| be |list|'s <a for=list>size</a>.
+
+<li><p>Remove all items from the list that match a given condition, or do nothing if none do.
+
+<li><p>Return true if |list|'s <a for=list>size</a> is less than |sizeBefore|; otherwise false.
```
zcorpan:
```suggestion
 <li><p>Let |sizeBefore| be |list|'s <a for=list>size</a>.

 <li><p>Remove all items from the list that match a given condition, or do nothing if none do.

 <li><p>Return true if |list|'s <a for=list>size</a> is less than |sizeBefore|; otherwise false.
```

zcorpan (approved):
LGTM % indentation

noamr:
> LGTM % indentation

@annevk are we happy with this :) ?

infra.bs
```diff
@@ -1626,8 +1626,17 @@ below for <a for=set lt=append>ordered set append</a>, <a for=set>prepend</a>, a
 index is to add the given item to the list between the given index &minus; 1 and the given index. If
 the given index is 0, then <a for=list>prepend</a> the given item to the list.
 
-<p>To <dfn export for=list,set>remove</dfn> zero or more <a for=list>items</a> from a <a>list</a> is
-to remove all items from the list that match a given condition, or do nothing if none do.
+<div algorithm=remove>
```
annevk:
```suggestion
<div algorithm>
```

infra.bs
```diff
@@ -1683,6 +1692,26 @@ following steps. They return a <a>list</a> of the same designation as |list|.
 </ol>
 </div>
 
+<div algorithm=has-duplicates>
```
annevk:
```suggestion
<div algorithm>
```

infra.bs
```diff
@@ -1683,6 +1692,26 @@ following steps. They return a <a>list</a> of the same designation as |list|.
...
+To check if a <a>list</a> |list| <dfn export for=list>has duplicates</dfn>:
```
annevk:
```suggestion
<p>To check if a <a>list</a> |list| <dfn export for=list>has duplicates</dfn>:
```

infra.bs
```diff
@@ -1683,6 +1692,26 @@ following steps. They return a <a>list</a> of the same designation as |list|.
...
+
+<ol>
+ <li><p>Let <var>seenItems</var> be « ».
+
+ <li>
+  <p><a for=set>For each</a> |item| of <var>list</var>:
+
+  <ol>
+   <li><p>If <var>seenItems</var> <a for=list>contains</a> |item|, return true.</p></li>
```
annevk:
```suggestion
   <li><p>If <var>seenItems</var> <a for=list>contains</a> |item|, then return true.</p></li>
```

annevk (approved):
Looks good modulo nits.

XavierMP14 (commented):
X

PR #710, opened 2026-05-08
Define identity/equality operators

### LINKED ISSUE #664: Equality definition rough plan?
### What is the issue with the Infra Standard?

See: https://github.com/whatwg/infra/issues/230, https://github.com/whatwg/infra/issues/643, https://github.com/whatwg/infra/pull/660, https://github.com/whatwg/html/pull/10924, https://github.com/whatwg/html/issues/10315

I think the following would work:

- Define "equals" for all of Infra's primitive data types. Most are "obvious". Special notes:
  - Infra doesn't really define numbers, but we should call out that we intend mathematical equality by default
  - We already have a definition of "is" for strings
- Disallow using "equals" on lists and maps, since it is too confusing.
- Not 100% sure on the right path for structs, but my initial proposal:
  - Say that it works by default by comparing itemwise
  - Specifications can override the definition of "equals" for their particular struct (e.g. [as URL does](https://url.spec.whatwg.org/#url-equivalence))
- Say that every other type we use in specs defaults to "reference equality", but can define their custom equals overload if they want.
  - Examples of such types that are not just structs in disguise: algorithm steps, IDL values, JS values
- With "equals" now defined for all types except lists and maps and things containing them, define "deep equals" for lists and maps, which recursively uses "deep equals" on keys/values/items
  - This can replace https://infra.spec.whatwg.org/#set-equal
- Optionally define "shallow equals" for sets and maps, although I'm not sure we have any use cases
- Go through existing places in Infra that use ambiguous notions of equality, like our definition of "list contains", and replace them to use "equals".

noamr:
- `Identical to` (or `is`) is for primitives or by-reference comparisons
- `Equal` is for shallow comparison of structs/lists/maps or custom data types, and is used by internal map/lists algorithms such as `list/contains`. It checks for cyclical references internally, using a private-type visited list.
- Primitive data types specify their own "considered identical" overload.
- Data types must specify "considered equal" in order to participate in list-contains or be map keys.

Closes #664

***
<a href="https://whatpr.org/infra/710.html" title="Last updated on Jul 6, 2026, 1:29 PM UTC (b3e4576)">Preview</a> | <a href="https://whatpr.org/infra/710/97e62b3...b3e4576.html" title="Last updated on Jul 6, 2026, 1:29 PM UTC (b3e4576)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -2110,8 +2110,8 @@ a <a>map</a> |map|, with a less than algorithm |lessThanAlgo|, is to create a ne
  </ol>
 </div>
 
-<p>Two <a>structs</a> are <a>considered equal</a> given |context| if all of their corresponding <a
-for=struct>items</a> are <a for=/>equal</a> given |context|.
+<p>Two <a>structs</a> are <a>considered equal</a> given |context| if all of their corresponding
+<a for=struct>items</a> are <a for=/>equal</a> given |context|.
 
 <h4 id=tuples>Tuples</h4>
 
--- a/infra.bs
+++ b/infra.bs
@@ -706,6 +706,47 @@ Standard that should be reported and addressed.
  </ol>
 </div>
 
+<h3 id=equality>Equality</h3>
+
+<p>Two values the same data type can be <a>identical</a>, <dfn export lt="equal|equals|shallow
+equal">equal</dfn>, or <dfn export lt="equivalent|deep equal">equivalent</dfn>.
+
+<p>Variables or references are <dfn export lt=is|identical>identical</dfn> if they refer to the same
+value, or if they are both of the same primitive type, and that primitive type's definition of
+<a>identical</a> applies to them.
+
+<p>Unless stated otherwise, two values of the same type are <a for=/>equal</a> if they are also
+<a>identical</a>. A data type can be specified to relax that condition and have its own definition
+of <a for=/>equal</a>.
+
+<p>Unless stated otherwise, two values of the same type are <a>equivalent</a> if they are also <a
+for=/>equal</a>. A data type can be specified to relax that condition and have its own definition of
+<a>equivalent</a>.
+
+
+<div class=example id=example-struct-equality>
+ <ol>
+  <p>A <dfn ignore>keyValue</dfn> is an example <a>tuple</a> consisting of a <dfn ignore>key</dfn>
+  (a string) and <dfn ignore>value</dfn> (a <a for=/>list</a> of numbers).
+
+  <li><p>Let |original| be the keyValue ("x", « 1, 2, 3, 4 »). <li><p>Let |alsoOriginal| be
+  |original|. <li><p>Let |copyWithSharedList| be the keyValue ("x", |original|'s value). <li><p>Let
+  |clone| be ("x", « 1, 2, 3, 4 »). <li><p><a>Assert</a>: |original| <a for=/>is</a> |alsoOriginal|
+  (they are <a>identical</a>). <li><p><a>Assert</a>: |original| <a for=/>is</a> not
+  |copyWithSharedList|. <li><p><a>Assert</a>: |original| <a for=/>is</a> not |clone| (they are not
+  <a>identical</a>). <li><p><a>Assert</a>: |original| is <a for=/>equal</a> to |copyWithSharedList|.
+  <li><p><a>Assert</a>: |original| is not <a for=/>equal</a> to |clone|. <li><p><a>Assert</a>:
+  |original| is <a>equivalent</a> to |clone|.
+ </ol>
+</div>
+
+<div class=example id=example-list-equivalence>
+ <ol>
+  <li><p>Let |x| be « 1, « 2, 3 », 4 ». <li><p>Let |y| be « 1, « 2, 3 », 4 ». <li><p><a>Assert</a>:
+  |x| and |y| are not <a>identical</a>. <li><p><a>Assert</a>: |x| and |y| are not <a
+  for=/>equal</a>. <li><p><a>Assert</a>: |x| and |y| are <a>equivalent</a>.
+ </ol>
+</div>
 
 <h2 id=primitive-data-types>Primitive data types</h2>
 
@@ -718,6 +759,7 @@ JavaScript <b>null</b> value. [[!ECMA-262]]
 
 <p class=example id=example-null-return>If <var>input</var> is the empty string, then return null.
 
+<p>Null values are always <a>identical</a>.
 
 <h3 id=booleans>Booleans</h3>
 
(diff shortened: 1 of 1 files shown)


annevk (commented):
Do we really need three types? I don't understand why we can't define equals for lists. I also think that giving "equivalent" a normative meaning will be quite awkward as we often use it in notes and such to mean something else.

I also somewhat strongly prefer "is" over "identical". As it becomes more common having to scan over the long form will be annoying.

noamr:
> Do we really need three types? I don't understand why we can't define equals for lists. I also think that giving "equivalent" a normative meaning will be quite awkward as we often use it in notes and such to mean something else.

I was going by the discourse on the issue... I am not sure if deep-equals/equivalent is necessary or if it currently would be used by anyone. I am ok with dropping it for the time being.

Sure we can define equals for lists/maps etc, I was going by the discourse.

> 
> I also somewhat strongly prefer "is" over "identical". As it becomes more common having to scan over the long form will be annoying.

"Identical to" was already defined for strings and in some sentences it fits more naturally, as in "a and b are identical". Both forms are exported so perhaps it's ok to leave the choice to embedders?

annevk:
Okay, thinking about this some more I think we want to borrow more of the original issue's ideas:

- For identity comparison we can reuse the terminology is/identical to from strings.
- Separate defining equality for something from the equals operator. So that you can say struct's equality is unordered deep equals, for instance.
- Use deep equals and unordered deep equals instead of equivalent.
- We should assert if we are comparing something with equals/deep equals and one of the underlying items has no equality.

I'm not sure how we should deal with cyclic references or whether lists/maps/structs should have a default equality that's not null. Perhaps we should start out with them not having a default and asserting until we get a bit more comfortable with it.

noamr:
> Okay, thinking about this some more I think we want to borrow more of the original issue's ideas:
> 
> - For identity comparison we can reuse the terminology is/identical to from strings.
> - Separate defining equality for something from the equals operator. So that you can say struct's equality is unordered deep equals, for instance.
> - Use deep equals and unordered deep equals instead of equivalent.
> - We should assert if we are comparing something with equals/deep equals and one of the underlying items has no equality.

This seems unnecessary. It should fall back to shallow-equals/identity. E.g a map of map of structs where one of the struct items is a platform object - it's ok if those are compared by reference and everything else is deep-equals.
> 
> I'm not sure how we should deal with cyclic references or whether lists/maps/structs should have a default equality that's not null. Perhaps we should start out with them not having a default and asserting until we get a bit more comfortable with it.

annevk:
For platform objects it seems okay to default (or enforce?) all of them to identity for their equality, but that should probably be in Web IDL.

noamr:
> For platform objects it seems okay to default (or enforce?) all of them to identity for their equality, but that should probably be in Web IDL.

Not sure I like this model. Maybe it is better to defer deep equality for now? Not sure if it has existing users?

annevk:
I think Sanitizer essentially wants something like that? Not having equals for lists or maps at all probably won't work.

noamr:
> I think Sanitizer essentially wants something like that? Not having equals for lists or maps at all probably won't work.

Sanitizer just needs to define equality for the SanitizerConfig dictionaries such as SanitizerElement, and then all the map/list algorithms (mainly contains) should refer to that. It doesn't need to deeply compare entire lists.

annevk:
I guess we don't have to define (unordered) deep equals for v1, but I think the point around equals and equality still stands. And so we'd still have to define somewhere what equality means for platform objects since I don't think we want it to default to is.

noamr:
> I guess we don't have to define (unordered) deep equals for v1, but I think the point around equals and equality still stands. And so we'd still have to define somewhere what equality means for platform objects since I don't think we want it to default to is.

In what scenario wouldn't we want to default to 'is'? Looks to me like unless there is a special equality operator like in URLs or structs it's probably the natural fallback?
Eg for js object references it's probably the most useful default.

noamr:
> > I guess we don't have to define (unordered) deep equals for v1, but I think the point around equals and equality still stands. And so we'd still have to define somewhere what equality means for platform objects since I don't think we want it to default to is.
> 
> In what scenario wouldn't we want to default to 'is'? Looks to me like unless there is a special equality operator like in URLs or structs it's probably the natural fallback? Eg for js object references it's probably the most useful default.

nevermind, I've looked at this again and was convinced. See new revision.

annevk (commented):
I still miss a clear distinction between how you define equality for a concept and then how you compare it. Currently we do things like `<dfn for=url>equals</dfn>` but maybe it should actually be something like

> A URL's equality, given a URL _a_ and a URL _b_, is:

or maybe the `<dfn>` pattern is okay, but I don't see you employing that here for list and set (although you do for unordered).

Also for sets the unordered definition is the same as the ordered one, which seems wrong.

noamr:
> I still miss a clear distinction between how you define equality for a concept and then how you compare it. Currently we do things like `<dfn for=url>equals</dfn>` but maybe it should actually be something like
> 
> > A URL's equality, given a URL _a_ and a URL _b_, is:

Yes, we need to add that to the URL spec.  Thinking to do there what we do here for strings, where there is still a `<dfn for>` for compat but it is also defined in terms of the "global" equality.

> 
> or maybe the `<dfn>` pattern is okay, but I don't see you employing that here for list and set (although you do for unordered).


I think you missed this line, which accounts for both `list` and `set`:

A [list](https://whatpr.org/infra/710.html#list) A is [equal](https://whatpr.org/infra/710.html#equal) to a [list](https://whatpr.org/infra/710.html#list) B if A has the same size as B and A[i] is [equal](https://whatpr.org/infra/710.html#equal) to B[i] for all [indices](https://whatpr.org/infra/710.html#list-get-the-indices) i of A.

> Also for sets the unordered definition is the same as the ordered one, which seems wrong.

Removed the extra one.

infra.bs
```diff
@@ -706,6 +706,44 @@ Standard that should be reported and addressed.
...
 </div>
 
+<h3 id=equality>Equality</h3>
+
+<p>Two values the same data type can be <a for=/>identical to</a> each other or <dfn export
+lt="equal|equals|shallow equal">equal</dfn>.
+
+<p>Variables or references are <dfn export>identical to</dfn> each other if they refer to the same
+value, or if they are both of the same primitive type, and that primitive type's definition of <a
+for=/>identical to</a> applies to them. A variable <dfn export>is</dfn> another variable when theh
```
aphillips: Typo.

```suggestion
for=/>identical to</a> applies to them. A variable <dfn export>is</dfn> another variable when they
```

noamr:
I think this is ready for another review?

JakobJingleheimer:
Heads up: there is a similar/related proposal going in TC39 ([`comparisons`](https://github.com/tc39/proposal-comparisons)).

PR #713, opened 2026-06-08
Explain literal syntax in terms of code points

### LINKED ISSUE #712: "surrounding by" vs "surrounding with" and clarity of `«»` vs `<< >>`
https://github.com/whatwg/infra/blob/97e62b3ec7cfee4cfae8c3e79bb9f3bf8aa0e0fc/infra.bs#L1553-L1556

1. I struggle with `by surrounding the list by « » characters`. Though using `by` with the past tense `surrounded` is common in US English (I don't know about UK or other non-US English) phrasing, `with` is more commonly used with the imperative `surrounding`.

2. `« »` is not certain to be understood to mean the European quotation chevrons, officially called "guillemets", and named "Left-Pointing Double Angle Quotation Mark" (`U+00AB` or `&laquo;`) and "Right-Pointing Double Angle Quotation Mark" (`U+00BB` or `&laquo;`) in Unicode and HTML.

   These are too often typos (or read as such) for `<< >>`, i.e., doubled angle brackets. I strongly suggest and request that the symbol names be included in at least the first instance, and preferably all instances. W3C has a specific [recommended presentation of such Unicode characters](https://www.w3.org/guide/manual-of-style/#Unicode) which I think would be well used here, but I would be probably happy enough with WHATWG's recommended presentation, if there is one.

   (I more strongly suggest and request that WHATWG include character names and Unicode numbers, and optionally HTML escapes, in all cases where such specific characters are required (e.g., `square brackets` in this paragraph) in this and other WHATWG standards. Misunderstanding of non-standard character naming and/or visual presentation in some fonts/browsers/etc. is too easy even if not frequent.)

annevk:
Fixes #712.

***
<a href="https://whatpr.org/infra/713.html" title="Last updated on Jun 10, 2026, 4:30 PM UTC (9500b7b)">Preview</a> | <a href="https://whatpr.org/infra/713/97e62b3...9500b7b.html" title="Last updated on Jun 10, 2026, 4:30 PM UTC (9500b7b)">Diff</a>

### DIFF
--- a/infra.bs
+++ b/infra.bs
@@ -2012,7 +2012,7 @@ a <a>map</a> |map|, with a less than algorithm |lessThanAlgo|, is to create a ne
 
 <p>A <dfn export>tuple</dfn> is a <a>struct</a> whose <a for=struct>items</a> are ordered. For
 notational convenience, a literal syntax can be used to express <a>tuples</a>, by surrounding the
-tuple with parenthesis and separating its <a for=tuple>items</a> with U+002C (,). To use this
+tuple with parentheses and separating its <a for=tuple>items</a> with U+002C (,). To use this
 notation, the <a for=tuple>names</a> need to be clear from context. This can be done by preceding
 the first instance with the name given to the <a>tuple</a>. An indexing syntax can be used by
 providing a zero-based index into a <a>tuple</a> inside square brackets. The index cannot be
--- a/infra.bs
+++ b/infra.bs
@@ -481,7 +481,7 @@ Variables must not be declared more than once per algorithm.
 
 <p>A multiple assignment syntax can be used to assign multiple variables to the <a>tuple</a>'s
 <a for=tuple>items</a>, by surrounding the variable names with U+0028 LEFT PARENTHESIS and
-U+0029 RIGHT PARENTHESIS, and separating each variable name by U+002C (,). The number of variables
+U+0029 RIGHT PARENTHESIS, and separating each variable name with U+002C (,). The number of variables
 assigned cannot differ from the number of <a for=tuple>items</a> in the <a>tuple</a>.
 
 <div class=example id=example-tuple-multiple-assignment>
--- a/infra.bs
+++ b/infra.bs
@@ -480,9 +480,9 @@ behavior; otherwise null.
 Variables must not be declared more than once per algorithm.
 
 <p>A multiple assignment syntax can be used to assign multiple variables to the <a>tuple</a>'s
-<a for=tuple>items</a>, by surrounding the variable names with parenthesis and separating each
-variable name by a comma. The number of variables assigned cannot differ from the number of
-<a for=tuple>items</a> in the <a>tuple</a>.
+<a for=tuple>items</a>, by surrounding the variable names with U+0028 LEFT PARENTHESIS and
+U+0029 RIGHT PARENTHESIS, and separating each variable name by U+002C (,). The number of variables
+assigned cannot differ from the number of <a for=tuple>items</a> in the <a>tuple</a>.
 
 <div class=example id=example-tuple-multiple-assignment>
  <ol>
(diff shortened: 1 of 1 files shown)


infra.bs
```diff
@@ -2010,10 +2012,11 @@ a <a>map</a> |map|, with a less than algorithm |lessThanAlgo|, is to create a ne
 
 <p>A <dfn export>tuple</dfn> is a <a>struct</a> whose <a for=struct>items</a> are ordered. For
 notational convenience, a literal syntax can be used to express <a>tuples</a>, by surrounding the
-tuple with parenthesis and separating its <a for=tuple>items</a> with a comma. To use this notation,
-the <a for=tuple>names</a> need to be clear from context. This can be done by preceding the first
-instance with the name given to the <a>tuple</a>. An indexing syntax can be used by providing a
-zero-based index into a <a>tuple</a> inside square brackets. The index cannot be out-of-bounds.
+tuple with parenthesis and separating its <a for=tuple>items</a> with U+002C (,). To use this
```
TallTed: Can't surround a tuple with one parenthesis. Need a pair of parentheses.
```suggestion
tuple with parentheses and separating its <a for=tuple>items</a> with U+002C (,). To use this
```

infra.bs
```diff
@@ -480,9 +480,9 @@ behavior; otherwise null.
 Variables must not be declared more than once per algorithm.
 
 <p>A multiple assignment syntax can be used to assign multiple variables to the <a>tuple</a>'s
-<a for=tuple>items</a>, by surrounding the variable names with parenthesis and separating each
-variable name by a comma. The number of variables assigned cannot differ from the number of
-<a for=tuple>items</a> in the <a>tuple</a>.
+<a for=tuple>items</a>, by surrounding the variable names with U+0028 LEFT PARENTHESIS and
+U+0029 RIGHT PARENTHESIS, and separating each variable name by U+002C (,). The number of variables
```
TallTed: I do not understand why it is apparently found better to *sometimes* present the full name of a character (`LEFT PARENTHESIS`, `RIGHT PARENTHESIS`), and *sometimes* present the character itself (`,`, `«`, `»`), even if they all provide the Unicode identifier (`U+0028`, `U+002C`, `U+0029`). This seems very likely to lead to misunderstanding, and to be perceived by the reader as in error. Issues are almost certain to arise when using wrappers such as braces `{}`, (square) brackets `[]`, and angle brackets `<>`, which names are often confused with each other or even displaced by another wrong name.
```suggestion
U+0029 RIGHT PARENTHESIS, and separating each variable name with U+002C (,). The number of variables
```

annevk: This is explained in the second paragraph of https://infra.spec.whatwg.org/#code-points

TallTed: "Documents are allowed to use "U+1F914 THINKING FACE (🤔)" as well, though this is somewhat verbose." is hardly an explanation.

Or did you mean, "U+0029 can be referred to as "U+0029 RIGHT PARENTHESIS", because even though it renders, this avoids unmatched parentheses."?

Unmatched parentheses are better avoided by editing and proofreading with care than by surrendering to careless efforts.

Rather than an argument in favor of leaving out the example characters, I think this is an argument in favor of more obvious styling, such as --

- A multiple assignment syntax can be used to assign multiple variables to the <a>tuple</a>'s <a for=tuple>items</a>, by surrounding the variable names with `U+0028` `LEFT PARENTHESIS` (`(`) and `U+0029` `RIGHT PARENTHESIS` (`)`), and separating each variable name with a `U+002C` `COMMA` (`,`). The number of variables assigned cannot differ from the number of <a for=tuple>items</a> in the <a>tuple</a>.

- For notational convenience, a literal syntax can be used to express <a>lists</a>, by surrounding the list with `U+00AB` `LEFT-POINTING DOUBLE ANGLE QUOTATION MARK ` (`«`) and `U+00BB` `RIGHT-POINTING DOUBLE ANGLE QUOTATION MARK` (`»`), and separating its <a for=list>items</a> with `U+002C` `COMMA` (`,`). An indexing syntax can be used by providing a zero-based index into a list inside square brackets (`[` and `]`). The index cannot be out-of-bounds, except when used with <a for=list>exists</a>.

Styling gets better and more obvious when CSS similar to that used by W3C is in play.

annevk: I'm not sure I'd want to mark it up as code or something else. That would not be how we'd render these characters normally and also doesn't always match their function. Anyway, if you feel strongly we can continue that discussion in a separate issue.

noamr (approved):

